PR #11, opened 2022-05-09
Features (Image / opacity / padding)

srush:
Adds a couple more features that needed from diagrams: 

* `image` shape for including images (from diagrams)
* opacity - for svg command
* `scale_x` functions to scale to size (from diagrams)
* `pad_x` functions that extend bounding box (from diagrams)

This example includes a png as an image, scales it down to unit size, grids it, draws an opaque box, and uses `pad_t` for spacing. 

image


danoneata (approved):
Thanks! Looks good to me 👍

PR #15, opened 2022-05-10
Two new examples: lenet and sunflower logo

srush:
Two more examples, play a bit with arrows and bounding boxes.


danoneata:
Very nice, thank you!

Issue #14, opened 2022-05-10
Multiple interpretations


danoneata:
Haskell's `diagrams` allows for multiple interpretations of the domain specific language (DSL). For example, the code
```haskell
mconcat . iterateN 3 (rotateBy (1/9)) $ triangle 1
```
can yield various outputs depending on how it is interpreted: the image on left, if interpreted as a diagram, or the third image, if interpreted as a trail.
![image](https://user-images.githubusercontent.com/819256/167637723-6f6bc4eb-cca4-4cd6-8316-60f4dcefa737.png)
(See [their manual](https://diagrams.github.io/doc/manual.html#the-traillike-class) for more information.)

This feature might be nice to have (it will allow drawing something like the [Hilbert curve](https://diagrams.github.io/gallery/Hilbert.html)), but I expect to be rather involved.

srush:
I honestly could not figure this part of the docs out... It seemed cool, but maybe a bit too cool?

danoneata:
Yes, I think you are right, it might be too confusing. If I really want that behavior, I guess I could have another data type that behaves differently. Specifically for paths (what is called `Trail`s in `diagrams`), I could potentially introduce a data type whose composition glues the end of a path to the start of the other (instead of overlapping them at their origins).

srush:
Chalk now roughly has this. It's a bit more explicit in python since we don't have the ability of the type system to disambiguate. But that feels pretty close to what we are going to get. 

If we wanted we could make the "monoid" structure a bit more explicit. (maybe having a general associative mconcat operator). But we can't do magic haskell stuff like have `empty()` work for all the types.

PR #18, opened 2022-05-11
Two more minor features. 

srush:
This PR adds two more small things that came up. 

First an `above` function notated as `//` that puts text above while keeping the origin of the lower box. This makes it way easier for me to add labels to my diagrams while keeping things centered. For example below the text is `//` over the arrows.  

image

Second a `Trail` class. This is a super stripped down version of the Haskell one that keeps a "translational-invariant" path. The monoid here is just a list (haskell uses something called a finger tree?), but it lets your apply transformations to the offsets.  At first I thought this datastructure was overkill, but it lets you do things that I couldn't do with diagrams, for example keep a running trail of points. (future todo is to figure out how this relates to Paths and Points, and whether transforms can be abstracted to apply to all of them.) 

image


danoneata:
Thanks!

Regarding the `Trail` data type, [that's exactly what I had in mind](https://github.com/danoneata/chalk/issues/14#issuecomment-1123732879)!

I've made a few changes:
- distinguish between `Point`s and `Vector`s
- added a few helper methods to `Trail` (such as, `rotate`, `scale`, `reflect_x`, `reflect_y`)
- added a new example based on the Koch curve

srush:
Oh neat. Yes at first I didn't realize these were so critical to diagrams, but reading the docs it does seems critical.

This also makes it easy to add `atPoints` for trails from paths/polygons, which I think you mentioned as well.

Did you add your code? I don't see it in the PR

danoneata:
Yes, I've added it; it's already [on the master](https://github.com/danoneata/chalk/commits/master). I didn't know how to modify your PR; is there an easy way to do that?

Issue #6, opened 2022-05-07
Support curves


srush:
Connectors need curves to look good.

Not sure how to do this. I like the "bend right" / "bend left" in tikz, but maybe Diagrams has a better API.

PR #20, opened 2022-05-12
Adds arcs

danoneata:
Nothing too fancy, just adds arcs; these could be useful for bendy arrows, as discussed in #6.

image

```python
import random

from itertools import cycle, product

from colour import Color
from chalk import *

colors = cycle("red orange blue".split())
circles = [
    circle(0.1).translate(i, j).fill_color(Color(c))
    for c, (i, j) in zip(colors, product(range(4), range(4)))
]

def random_arc():
    r = lambda: random.randint(0, 3)
    h = random.random()
    return arc_between((r(), r()), (r(), r()), h)

dia = concat([random_arc() for _ in range(10)] + circles)
dia.render("/tmp/o.png", 256)
dia.render_svg("/tmp/o.svg", 256)
```


srush:
Apparently the type of transformable is messing up mypy a bit. I will fix this in a separate PR `shape.rotate(v.angle).translate(p.x, p.y)`. This is a known issue with the python type system https://peps.python.org/pep-0673/

I think probably worth checking in the example too. More examples seem good. My suggestion would be to fix the random seed. Then we can add a test that new changes don't break the SVG of old examples

PR #21, opened 2022-05-12
Fun with vectors and paths 

srush:
having a vector type opens up some neat diagrams potential. This example draws cubes to illustrate tensor broadcasting. This is hard to do with the diagram object directly since bounding boxes are squares. However with paths can compose smaller cubes and then compose the whole diagram. 

image


danoneata:
Very cool example! 👍 

I believe that the cube could also be drawn as follows (I've added the `shear_{x,y}` and `scale_{x,y}` transformations to the PR):
```python
face_m = rectangle(1, 1)
face_t = rectangle(1, 1).scale_y(0.5).shear_x(-1)
face_r = rectangle(1, 1).scale_x(0.5).shear_y(-1)
```

It might be also worthwhile augmenting the concatenation functions (`hcat`, `vcat`) with an additional `sep` parameter ([as in diagrams](https://hackage.haskell.org/package/diagrams-lib-1.4.5.1/docs/Diagrams-Combinators.html#v:cat-39-)), which would specify how much white space to insert between the adjacent diagrams.

chalk/__init__.py
```diff
@@ -5,10 +5,10 @@
 
 from chalk.core import Diagram, Empty, Primitive
 from chalk.shape import Circle, Rectangle, Path, Text, Image
-from chalk.point import Point
+from chalk.point import Point, Vector
 from chalk.trail import Trail
 
-ignore = [Trail]
+ignore = [Trail, Vector]
```
danoneata, line 11: I was curious: how is this `ignore` variable used?

srush: Oh I just added that to make `flake` not throw an error, but have Vector available to the user.

srush:
Oh `shear` is so cool. was thinking there was a better way, but I couldn't figure it out.

srush:
Some major changes:

* Switch to shear. 
* Add a Spacer class and struts to __init__
* Add a spacer argument to hcat and vcat
* Implement a `Transformable` class for all the objects that can be transformed with the same API.

chalk/__init__.py
danoneata: Ah, I see, thanks for the explanation!

danoneata:
Thanks a lot! Very nice contributions!

PR #25, opened 2022-05-13
Add latex rendering

srush:
Renders latex through latextools. This library produces a SVG of the latex output. In svg mode this can be rendered and transformed directly as an element. Currently not supported in Cairo, but in theory we could add it as an image. 

Downside: brings in some dependencies. maybe these could be made optional. 

(After this I need to get back to my day job, too much fun. )


danoneata:
Awesome! I'll play with the PR these days!

> (After this I need to get back to my day job, too much fun.)

Ha-ha! Your productivity was amazing! I was short on time these days, otherwise I would have loved (to try) to keep up with you 🙂

examples/latex.py
```diff
@@ -0,0 +1,20 @@
...
+
+def label(text):
+    return latex(text).scale(0.5).pad_b(0.4)
+
+def arrow(text, d=True):
+    return label(text) // left_arrow
+
+# Autograd 1
+d = hcat([arrow(r"$f'_x(g(x))$"), box("$f$"), arrow(r"$f'_{g(x)}(g(x))$"), box("$g$"), arrow("1")], 0.2)
+d.render_svg("examples/output/latex.svg", 100)
```
danoneata, line 20: Could you please commit the rendered SVG file to the repo?

PR #27, opened 2022-05-20
Define a default renderer for Ipython notebooks

srush:
Defines a default renderer so that you can easily play with diagrams in Jupyter / Colab notebooks. 

image

Also converts the readme file to be rendered file through jupytext. I think this is a preferable option to streamlit, since you can edit in streamlit if you want or just as a regular python file. This is also a pretty easy way to convert the examples to literate markdown files with inline rendering.


srush:
Only downside I see with this approach is that it doesn't take a "height" option. Maybe we can have a special diagram wrapper that defines a global height.

Makefile
```diff
@@ -47,3 +47,5 @@ hex_variation:
 
 images: squares hanoi intro escher_square lenet logo hilbert koch tensor latex hex_variation
 
+README.md: readme.py
+	jupytext --to ipynb --execute readme.py; jupyter nbconvert --to markdown --output README.md readme.ipynb
```
danoneata, line 51: Should this have been
```bash
jupytext --to ipynb --execute README.md
```
Or did you just forget to commit the `readme.py` file? Or am I misunderstanding something?

srush:
I forgot to a add... Here it is with the readme.py.

danoneata:
Having the diagrams automatically displayed in an IPython notebook is a very nice feature! Thank you!

I also really like the idea of bundling together the code and its graphical output in a literate markdown file.

I'm not yet convinced that we should have the `README.md` file automatically generated, since in certain cases I would like to have more fine-grained control over what's in there. For example, the generated image paths are relative, but it would be easier if they were absolute paths, since otherwise [they are not displayed in PyPI](https://pypi.org/project/chalk-diagrams/0.1.dev0/). Also I'm not a big fan of the default naming convention used by `nbconvert`—do you know if it's possible to change that? That said, it might be worthwhile to have the examples from the `README.md` included as an IPython or Colab notebook.

> Only downside I see with this approach is that it doesn't take a "height" option. Maybe we can have a special diagram wrapper that defines a global height.

I don't know if this is what you are proposing, but we can have a default render height associated with the `Diagram` class and this could be overwritten using a setter method, for example:
```python
circle(1).fill_color(blue).render_height(256)
```
Would something like this be ergonomic enough?

srush:
> I also really like the idea of bundling together the code and its graphical output in a literate markdown file—this change would make the examples much more readable.

Makes sense! I'll revert that part. (I will note that I found a bug when rendering it, which is a nice benefit)
 
> I'm not yet convinced that we should have the `README.md` file automatically generated, since in certain cases I would like to have more fine-grained control over what's in there. For example, the generated image paths are relative, but it would be easier if they were absolute paths, since otherwise [they are not displayed in PyPI](https://pypi.org/project/chalk-diagrams/0.1.dev0/). Also I'm not a big fan of the default naming convention used by `nbconvert`—do you know if it's possible to change that? That said, it might be worthwhile to have the examples from the `README.md` included as an IPython or Colab notebook.

Let me look into whether nbconvert can be configured to export absolute paths with better names, or even better render images inline. I agree with all your points, but I do really like the idea of having literate markdown examples with inline images to teach people how to use the library (and they could also have an "edit in colab" link. )


> ```python
> circle(1).fill_color(blue).render_height(256)
> ```

Yeah, this is the right frontend. but I guess it should probably be functional on the backend. Something like wrapping it in a `ToRender` class that can be removed.

srush:
removed.

danoneata:
>  but I do really like the idea of having literate markdown examples with inline images to teach people how to use the library (and they could also have an "edit in colab" link.)

I completely agree! I wonder if we should just have all the files in the `examples` folder as IPython notebooks: these could be rendered by Github ([example](https://github.com/danoneata/chalk/blob/danoneata/colab/examples/intro.ipynb)) and they are easy to open in Google Colab (by linking their URL; [example](https://colab.research.google.com/github/danoneata/chalk/blob/danoneata/colab/examples/intro.ipynb)). Or would you instead suggest to have multiple versions in the repo: the Python file, the IPython notebook and the corresponding Markdown file (with the last two others being automatically generated through `jupytext` and `nbconvert`)? I fine with either option.

srush:
For the literate code I would suggest we edit them as python files, in the `light` jupytext format and commit those. Then also commit the rendered .ipynb as output (but never edit the notebooks directly, that's for users to view/edit). 

We don't need to go through Markdown. If later we decide we want a website or docs we can mess with nbconvert to generate markdown or html.

I'll send a PR for the size thing. I think my only point is that I want to keep the code functional, and not modify a diagram directly.

srush:
(Committing and moving this discussion to issues).

PR #34, opened 2022-05-21
Make `Path` an instance of `Transformable` 

danoneata:
Allows the placement on diagrams on a _transformed_ path. For example

```python
from colour import Color
from chalk import *
from chalk.shape import Path

green = Color("green")
nodes = [circle(0.2).fill_color(green) for _ in range(6)]
path = Path.polygon(6, 1).shear_x(0.5)
place_on_path(nodes, path)
```

produces 

image

Other changes include:

- refactored the definition of some shapes (`hrule`, `vrule`, `polygon`) as `Path` (so they can be used for placement)
- updated the `Trail.from_path` constructor to take at input a `Path` instead of a `Diagram`-wrapped `Path`,
since the latter resolves the transformation at rendering and might lead unexpected results (e.g., `Trail.from_path(polygon(6, 1).shear_x(0.5))` wouldn't have taken the transformation into account).

### DIFF
--- a/chalk/__init__.py
+++ b/chalk/__init__.py
@@ -27,8 +27,7 @@ def empty() -> Diagram:
 def make_path(
     coords: List[Tuple[float, float]], arrow: bool = False
 ) -> Diagram:
-    points = [Point(x, y) for x, y in coords]
-    return Primitive.from_shape(Path(points, arrow))
+    return Primitive.from_shape(Path.from_list_of_tuples(coords, arrow))
 
 
 def circle(radius: float) -> Diagram:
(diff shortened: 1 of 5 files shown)


srush:
This is great. Seems like an improvement across the board. I particularly like unit_x and unit_y.

I guess the only questions are: 
* Should `make_path` return a diagram or a path?
* Should we make it clear which classes / functions are "public". Seems like Path and Trail are public, but not clear if Transform / Polygon etc should be part of the user API.

danoneata:
> I guess the only questions are:
> - Should `make_path` return a diagram or a path? 
> - Should we make it clear which classes / functions are "public". Seems like Path and Trail are public, but not clear if Transform / Polygon etc should be part of the user API.

Yes, these are valid concerns! I've also thought of the potential confusion that these changes might introduce.

My reasoning was to have `Diagram`-only constructors and combinators in the `chalk/__init__.py` file, since I would expect this to be the most common way of using the library. Constructors of `Path`s and `Trails`s could be available in their corresponding modules (maybe we can have `chalk.path` and `chalk.trail`); I guess these should also be "public," but it might be good to keep them separate from `Diagram`s given that their semantics are different. Regarding the `Transform` class, I don't know think it's necessary to be exposed, since it can be accessed as method calls on `Diagram` (or `Path` or `Trail`) objects.

PR #43, opened 2022-06-02
Adding diagram display utilities to `chalk`

sugatoray:
This PR adds the functionality of displaying a `chalk.Diagram` with a few convenience functions.

- [x] Adds the new `chalk.utils` module: Closes #44 

  - [x] `utils.show(filepath: str)`: 
    > Use this function for displaying saved images from disk.
  
  - [x] `utils.imgen(d: Diagram, **kwargs)`: 
    > Use this function to display a diagram, without creating any permanent file on disk. 
    > :point_right: This function does the following:
    > 1. Creates a temporary file.
    > 2. If necessary, it also creates a temporary folder for storing the temporary file.
    > 3. Displays the temporary-image-file for a user-specified (optional) `5` seconds.
    > 4. Clears up both the temporary file and folder after the 5 seconds have elapsed.

  - [x] `utils.quick_probe(**kwargs)`:
    > Use this function to **quickly generate a default sample Diagram** (or a user-provided Diagram) and **display it _without saving it_**.

  - [x] `create_sample_diagram()`:
    > Use this function to create a sample diagram, with multiple shapes, colors, etc. -- *with only a single function call*.

I believe, with these utilities it will be quite easy to visualize diagrams without having to save them on the disk.

- [x] Adds a `Diagram.display()` method: Closes #47
  > This facilitats convenient visualization of a diagram without saving any image file permanently on disk. 

- [x] Adds docstrings to methods of:
  - [x] `chalk.core.Diagram` class. Closes #46 
  - [x] `chalk.bounding_box.BoundingBox` class. Closes #48  
  - [x] `chalk.point.Point` class. Closes #50 
  - [x] `chalk.point.Vector` class. Closes #49 
  - [x] `chalk.trail.Trail` class. Closes #51 
  - [x] all the classes other than `Diagram` in `chalk/core.py`: Closes #52

Issue #42, opened 2022-05-31
Add a `dev_requirements.txt` file


sugatoray:
Add a `dev_requirements.txt` file:

- Add the following requirements to it, as these are not listed anywhere for quick dev-environment setup.
  - [x] `black`
  - [x] `flake8`
  - [x] `mypy`
  - [x] `interrogate` (necessary for #41)

:fire: This WILL NOT be source-tracked.

PR #53, opened 2022-06-06
Add documentation website with MkDocs

### LINKED ISSUE #54: Add a Documentation website for the library
Adding a docs website with MkDocs framework could yield very pretty results. Also, maintaining MkDocs documentation website takes minimal effort.

sugatoray:
Add a documentation website with MkDocs: Closes #54 

- https://github.com/danoneata/chalk/pull/54

Note: this builds on top of the changes suggested in:

- https://github.com/danoneata/chalk/pull/43

So, please merge this after merging #43. 

You can see a live version of the website here: https://sugatoray.github.io/chalk/

**Added/updated:**

- [x] dependencies for docs: `requirements/docs.txt`
- [x] dependencies for development: `requirements/dev.txt` 
  - [x] `isort`: sorting import statements
  - [x] `darglint`: checking arg types in docstring
  - [x] `pre-commit`: introduced for gradual adaption
- [x] Updated docs:
  - [x] In a number of modules.
- [x] Updated infrastructure:
  - [x] updated Makefile: added multiple commands
  - [x] updated workflow: 
    - [x] `.github/workflows/mkdocs-publish-ghpages.yml`
    - [x] `.github/workflows/checks.yml`


sugatoray:
## Preview of Website

<details>
<summary>Home Page</summary>
<p>

image

</p>
</details>

<details>
<summary>API Page</summary>
<p>

image

</p>
</details>

sugatoray:
cc: @danoneata @srush

srush:
Wow this is amazing. 

Dumb question. Is there a way to avoid double writing the types? Or will it enforce if they are out of sync automatically?

srush:
To answer my own question it looks like `darglint` checks that the code docs match the types. Could you add that to the makefile checks?

sugatoray:
Can you give me an example of double writing the types? (A screenshot)

srush:
I just mean that the types are in the code, in the functions  defs, but you also repeat them in the docstring. I'm okay if that is needed for MkDocs, but I want to ensure they always match.

sugatoray:
No the extra ones in the brackets are not necessary. So, will drop them in the other PR #42.

srush:
Hi @sugatoray .

No, sorry I disagree. This is exactly the right PR to run darglint. It will check that all the doc strings you are adding are in the right format, and that the extra types that are in the docstrings always match the types that are in the code.

sugatoray:
@srush Ok. Will include it then.

srush:
Thanks. This looks nice mostly nice. I have two questions: 

- Is it necessary to check in all the javascript code in `docs/`? It seems like a lot to have in the repo. Is that the style MkDocs encourages?

- Is there a way we can avoid documenting things that are vacuous descriptions? Much of the documentation is great, but I don't really see the point of documentation like this: 

Since we already have the type in comments I would prefer if instead we preferred: 

```
    @property
    def tr(self) -> Point:
        """Returns the position of the top-right corner
        of the bounding box.
        Returns:
            Point: A point object.
        """
```

be instead ->

```
@property
    def tr(self) -> Point:
        "Returns the position of the top-right corner of the bounding box."
```      

I believe that this is called `short` style  https://github.com/terrencepreilly/darglint#strictness-configuration

I don't think these need to be done this PR, but I do feel like non-descriptive docstrings can be worse than nothing.

srush:
One other comment about the doc website: 

Many of these functions are not "user facing" and I would rather they not be in the docs. The only functions the user should know about are the functions in `__init__` , a subset of the functions in `Diagram`. Is there a way in MkDocs to have a statement in the docstring that makes these functions hidden in the docs?

image

sugatoray:
> One other comment about the doc website:
> 
> Many of these classes / functions are not "user facing" and I would rather they not be in the docs. The only functions the user should know about are the functions in `__init__` , a subset of the functions in `Diagram`. Is there a way in MkDocs to have a statement in the docstring that makes these functions hidden in the docs?
> 
> image

There should be such options to selectively add API information. But I don't know exactly how.

But what about adding a statement for now, on which methods are user-facing?

sugatoray:
Hi @srush, 

> Thanks. This looks nice mostly nice. I have two questions:
> 
> * Is it necessary to check in all the javascript code in `docs/`? It seems like a lot to have in the repo. Is that the style MkDocs encourages?

I was experimenting with a few things. I believe the amount of JS code could be cut down.

> * Is there a way we can avoid documenting things that are vacuous descriptions? Much of the documentation is great, but I don't really see the point of documentation like this:
 
I was only creating a construct so that both you and @danoneata can start adding more meaningful comments to the docstrings, making the documentation quality more rich. I know there are many such methods, where the documentation is present only to show it up on the docs website. Now, that the website can be up and running, both of you can keep adding to the documentation at your own pace.

> Since we already have the type in comments I would prefer if instead we preferred:
> 
> ```
>     @property
>     def tr(self) -> Point:
>         """Returns the position of the top-right corner
>         of the bounding box.
>         Returns:
>             Point: A point object.
>         """
> ```
> 
> be instead ->
> 
> ```
> @property
>     def tr(self) -> Point:
>         "Returns the position of the top-right corner of the bounding box."
> ```
> 
> I believe that this is called `short` style [terrencepreilly/darglint#strictness-configuration](https://github.com/terrencepreilly/darglint#strictness-configuration)
> 
> I don't think these need to be done this PR, but I do feel like non-descriptive docstrings can be worse than nothing.

There are three different tools at play here:

1. `darglint`: The short docstring works here.
2. `Material-for-MkDocs`: This is the major doc-infrastructure.
3. `mkdocstring`: The MkDocs extension that helps you render the generated API or API-related other documentation. This requires the first way of writing docstring (following `google` style), so that it can create the documentation as you see it is now.

However, if you find it otherwise, and if there are ways to cut corners and still have the docs publish as it shows now, please let me know.

sugatoray:
> Not sure I understand. Where do these files come from?
> 
> [docs/api/style.md](https://github.com/danoneata/chalk/pull/53/files#diff-afef9923fe779d9229138604a9063b8928830917ab3b7533826b6e4752b2bd96)
> 
> Are they autogenerated? We probably should just not generate them for most classes.

No they are not auto generated. I made them. But the code written inside, helps auto-generate the API docs.

sugatoray:
@srush For what it's worth, I use a VS Code extension to quickly generate the stub of the docstring from the function definition. Please see this comment for more details on it.

- https://github.com/danoneata/chalk/pull/43#issuecomment-1146717252

srush:
Okay, I guess this is a fine place to start from in terms of docs. Probably will end up removing a lot of them though. 

Just to be clear though, one should not be touching the internals of Composite. From the users perspective they should not know the difference between the different subclasses of diagrams. Those method and classes are private.

README.md
```diff
@@ -1,3 +1,18 @@
+<!--- BADGES: START --->
+[![GitHub - License](https://img.shields.io/github/license/danoneata/chalk?logo=github&style=flat&color=green)][#github-license]
+[![PyPI - Python Version](https://img.shields.io/pypi/pyversions/chalk-diagrams?logo=pypi&style=flat&color=blue)][#pypi-package]
```
srush, line 3: Do we need all this stuff? Feels like noise.

README.md
```diff
@@ -11,24 +26,38 @@ Jeremy Gibbons's lecture notes on [Functional Programming for Domain−Specific
...
 ```
+
 For the LaTeX extension you might need to install `pdf2svg` and `texlive`;
 on Debian these dependencies can be installed as follows:
+
 ```bash
 sudo apt-get install pdf2svg texlive texlive-science texlive-latex-recommended texlive-latex-extra
 ```
 
+**Installation with Conda**
```
srush, line 53: Is this needed? I feel like Conda users know this.

Makefile
```diff
@@ -113,21 +148,80 @@ check: style type test clean
...
-	if [ -f $(REQ_FILE) ]; then python -m pip install -r $(REQ_FILE); fi
+# .PHONY: install.e
+install.e: clean
+	@echo "📀🟢🔵 Installing $(PACKAGE_NAME) from local source ... ⏳"
 	python -m pip install -Ue "."
 
-installextras: install
-	@echo "📀 Installing $(PACKAGE_NAME) extra-dependencies from PyPI ... ⏳"
-	if [ -f $(DEV_REQ_FILE) ]; then python -m pip install -r $(DEV_REQ_FILE); fi
+# .PHONY: install
```
srush, line 163: Is this stuff all needed? It feels like noise.

srush (approved):
Generally looks good. A couple small aspects I would like to remove.

Makefile
sugatoray: The make file commands are like your API. Sometimes they will also require some restructuring. Some will be for most people to use and some only for a select few. 

If you have a contribution guideline, you could add the make commands, effectively smoothening out any steps that require extra attention. 

But installation as a make command is always very helpful.

README.md
srush: I don't. I think it is redundant. Would prefer less prelude.

Makefile
srush: I would like you to remove these. I think probably the makefile was a mistake. It was meant to be small and easily readable. Now it has dozens of complex commands.

README.md
srush: It is not necessary because it adds visual noise to the top of the readme and provides very little information. If someone needs to know these things it is trivial for them to look it up.

srush:
@sugatoray I appreciate the PR, but I am finding it to be a lot of back and forth. If you want less comments, send a PR with smaller changes more focused changes.

srush:
Merging now, so I can fix up the docs on my plane flight.

PR #58, opened 2022-06-09
[For discussion] Attempt at documenting a public API

srush:
This PR attempts to document the `public` API of diagrams with the goal of reducing the footprint of the docs (and letting us keep changing internal things). I did this by writing a couple of python scripts that show off the main features with visual examples. 

image

I think I like this approach. It is easy to do, doesn't require manually running things, and works with examples. We could also make the `help` part a bit more pretty and remove the notebook-y display fomat. 

However the main downside is that you don't get the full docstring of every function. (and the docstrings look ugly). Alternative would be to have a separate collection of pretty doc strings and have this as a tutorial like section. 

image

This PR also removes some of the extra mkdocs extensions. Would like to get this down to the minimal amount we need, however I do not know what a lot of them do.

@sugatoray


mkdocs.yml
```diff
@@ -98,37 +98,12 @@ plugins:
...
-      lib_version: 0.11.5
-      view_version: 0.2.6
-  - kroki:
-      ServerURL: https://kroki.io
-      EnableBlockDiag: true
-      Enablebpmn: true
-      EnableExcalidraw: true
-      EnableMermaid: true
-      DownloadImages: false
-      DownloadDir: docs/assets/kroki_generated
```
sugatoray, line 126: :+1: I would have dropped `kroki` as well. But `markmap` gives you the ability of adding mindmaps with markdown for better explanation of steps.

I would suggest you keep it for now. Why? Well, you need sequential steps to produce a certain visualization. You can think of each intermediate output as a node on a mindmap, followed by an operation. A mindmap could be helpful here showing this.

- https://github.com/neatc0der/mkdocs-markmap
- https://markmap.js.org/
- example: https://markmap.js.org/repl

image

mkdocs.yml
```diff
@@ -269,15 +244,15 @@ markdown_extensions:
 
 
 ### Customization
-extra:
+# extra:
```
sugatoray, line 247: You can use variables from the global scope to reuse in anywhere in the documentation. Variables defined by the user under `extra` are available globally.

mkdocs.yml
```diff
@@ -269,15 +244,15 @@ markdown_extensions:
...
   #   provider: mike
   # disqus: !ENV REPO_OWNER
-  alternate:
-  - name: en - English
-    link: /
-    lang: en
+  #alternate:
+  #- name: en - English
+  #  link: /
+  #  lang: en
```
sugatoray, line 255: :+1: Since, the documentation is available only in English, this is not necessary for now. But if you want to make documentation available in other languages, then this setting could be helpful.

mkdocs.yml
```diff
@@ -332,29 +307,20 @@ extra_javascript:
...
+    - Style: api/style.py
+    - Trails: api/trails.py
+    - Connections: api/names.py
+    - Utils: api/utils.md
+  - Examples:
+    - Hanoi: examples/hanoi.py
+    - Koch: examples/koch.py
+    - LeNet: examples/lenet.py
+    - Squares: examples/squares.py
+    - Tensor: examples/tensor.py
```
sugatoray, line 323: :+1:

Makefile
```diff
@@ -242,38 +242,13 @@ pipinstalltest:
...
-	# if pregendocs.examples is included as well.
-
 ## Gendocs
 
 # .PHONY: gendocs
 gendocs:
 	@echo "🔥 Generate documentation with MkDocs ... ⏳"
 	# generate documentation
-	mkdocs serve
+	mkdocs serve --dirtyreload
```
sugatoray, line 251: `mkdocs serve` alone makes sure changes are picked up. There is a `watch` parameter in `mkdocs.yml` that keeps track of files in various folders. In case, it is not picking up a change, hitting a `ctrl + s` on any of the `.md` files inside `docs` folder will enforce a refresh.

chalk/core.py
```diff
@@ -146,19 +146,6 @@ def _repr_svg_(self) -> str:
...
-        shapes.
-
-        💡 ``a.atop(b)`` is equivalent to ``a + b``.
-
-        Args:
-            other (Diagram): Another diagram object.
-
-        Returns:
-            Diagram: A diagram object.
-        """
```
sugatoray, line 161: Why are you removing this? Is it incorrect? Since, `Diagram` class will be used everywhere, doesn't it make sense to provide docstrings here?

Having examples somewhere else could be helpful. But that should not make docstrings redundant. Shouldn't it?

chalk/core.py
```diff
@@ -167,18 +154,6 @@ def atop(self, other: "Diagram") -> "Diagram":
...
-        places ``b`` on the right side of ``a``. This moves
-        ``b`` up to touch ``a``.
-
-        💡 ``a.beside(b)`` is equivalent to ``a | b``.
-
-        Args:
-            other (Diagram): Another diagram object.
-
-        Returns:
-            Diagram: A diagram object.
```
sugatoray, line 180: Same as [this comment](https://github.com/danoneata/chalk/pull/58/files#r893055676).

chalk/core.py
```diff
@@ -189,18 +164,6 @@ def beside(self, other: "Diagram") -> "Diagram":
...
-        touch ``a``.
-
-        💡 ``a.above(b)`` is equivalent to ``a / b``.
-
-        Args:
-            other (Diagram): Another diagram object.
-
-        Returns:
-            Diagram: A diagram object.
-        """
```
sugatoray, line 203: Same as [this comment](https://github.com/danoneata/chalk/pull/58/files#r893055676).

Makefile
```diff
@@ -242,38 +242,13 @@ pipinstalltest:
...
-	cp -rf examples/* docs/examples/
-
-# .PHONY: pregendocs.local
-pregendocs.local: pregendocs.doc
-
-# .PHONY: pregendocs.remote
-pregendocs.remote: pregendocs.doc # pregendocs.examples
-	# It was observed that the build fails on github pages
-	# if pregendocs.examples is included as well.
-
```
sugatoray, line 269: :bulb: This is fine. But this will also require changing the GitHub actions workflow for mkdocs-publishing.

mkdocs.yml
sugatoray: For `kroki` you get the following:

kroki_cheatsheet_20210515_v1 1_EN-1

sugatoray:
@srush Also, it will be nice to see it in action. Could you please set it up with your GitHub pages so the docs are available to easily see the output?

If you need help with this, please let me know.

mkdocs.yml
srush: We can add these back if we find a compelling use case.

mkdocs.yml
srush: If that happens (unlikely) we can add it.

chalk/core.py
srush: These comments are really good, but I moved them to the version of this function in `__init__`. https://github.com/danoneata/chalk/pull/58/files#diff-ebd35dd9c16e8a4cb47f718f53b6c8e9754d5c7666711be9b750956aa5ba8b57R237

The reason is that I don't think users should use this function. Instead they should either do `a + b` or `atop(a, b)`. This isn't your issue, but I think there are currently too many ways to do some of the core commands.

srush:
Sure, I set it up here https://srush.github.io/pydiagrams/

sugatoray:
@srush Beautiful! They look terrific! 💯🚀🎉

sugatoray:
Also, could you please update the readme and remove **Installation with Conda** (bit of an unnecessary prelude), as you had pointed out earlier?

sugatoray:
> This PR also removes some of the extra mkdocs extensions. Would like to get this down to the minimal amount we need, however I do not know what a lot of them do.

Do you need any help with identifying which ones you could do away with?

PR #71, opened 2022-06-17
Tikz backend v1

srush:
First pass at a tikz backend for rendering PDF files. Kind of shocked this works, but here is Hanoi 

image

Will cleanup the rest and get it checked in.


danoneata:
> Kind of shocked this works, but here is Hanoi

Ha-ha! Very nice! I'll take a look at the code as well.

I was wondering: would it make sense and would it be possible to use TikZ's capabilities to style the arrows (e.g., select the head tips, slightly shorten the edges)?

srush:
Okay, this works now.  Here is a pdf comparing outputs. It's cool how sharp the fractals look in PDF form compared to PNG https://github.com/danoneata/chalk/blob/b8093f23b9192756a42f928ea9fa112c9b0e6170/examples/output/comparison.pdf

The main idea is to use the TikZ cm= command which is roughly the same as Cairo Matrix and svg "Matrix". (although annoyingly they all have different argument order!). To make these all work I switched to using the Affine Library. 

There are two issues with TikZ though.
* Line width isn't transformation invariant. 
* Text size is transformation invariant. (Fonts are a bit gross in tex.)

To make these work I need to pass scale information through the style class. I don't love this, but I think is okay to do for now. 

Yes I think Arrows in tikz are great! Way better than SVG. We probably can just use their system.

1) It doesn't aut

danoneata:
> Okay, this works now. Here is a pdf comparing outputs. It's cool how sharp the fractals look in PDF form compared to PNG https://github.com/danoneata/chalk/blob/b8093f23b9192756a42f928ea9fa112c9b0e6170/examples/output/comparison.pdf

Yup! I guess both PDF and SVG look sharp since they are rendered as vector graphics.

> The main idea is to use the TikZ cm= command which is roughly the same as Cairo Matrix and svg "Matrix". (although annoyingly they all have different argument order!). To make these all work I switched to using the Affine Library.

Yes, I think that it is a good idea to have an uniform interface for all these transformations and `Affine` seems to fit the bill (I also like that it allows to compute the inverse, which I need [for transforming traces](https://github.com/danoneata/chalk/blob/danoneata/trace/chalk/trace.py#L42)).

> There are two issues with TikZ though.
> * Line width isn't transformation invariant.
> * Text size is transformation invariant. (Fonts are a bit gross in tex.)

A few more points:
- The `render_tikz` is not implemented for `Text`; is there a particular difficulty in adding this?
- This is minor, but it seems to me that [the rendered PDF is not `pad`ed](https://github.com/danoneata/chalk/blob/cairo_drop/chalk/core.py#L151) (I assume because the `standalone` LaTeX package crops tightly around the drawn region?)
- I'm having some issues rendering this example as a PDF:
```python
from chalk import *

diagrams = [
    square(1).line_width(0.2),
    square(1).line_width(0.2).scale_x(2),
    square(1).line_width(0.2).scale(2),
    square(1).scale(2).line_width(0.2),
]

dia = hcat(diagrams, sep=0.5).pad(0.5)
dia.render_pdf("/tmp/o.pdf")
```
I'm getting [an incomplete image](https://github.com/danoneata/chalk/files/8935344/o.pdf).

> To make these work I need to pass scale information through the style class. I don't love this, but I think is okay to do for now.

I agree that this is a bit annoying; and I guess it will be difficult to get a completely consistent behaviour to the other backends, which scale the lines differently if a non-uniform transformation is applied! For example,
```python
diagrams = [
    square(1).line_width(0.2).scale_x(2),
    square(1).line_width(0.2).scale(2),
]
dia = hcat(diagrams, sep=0.5).pad(0.5)
```
yields
scaling-image

It can be argued if whether this is what we expect from the above code; from what I see [`diagrams` have dropped this behaviour](https://archives.haskell.org/projects.haskell.org/diagrams/doc/manual.html?ref=driverlayer.com/web#local-units) in favour of uniform scaling of lines even if the transformation is non-uniform:
 
>     1. It doesn't aut

It seems that this sentence got chopped half way through?

srush:
Thanks for the comments, very helpful. 

Here is the new updated PDF https://github.com/danoneata/chalk/blob/cairo_drop/examples/output/comparison.pdf . It contains: 

* Padding (done by adding an explicit Path spacer behind diagram)
* Text (Assumes sizes of sans-serif will be roughly proportional to the size as cairo) 
* Attempt at line width scaling. Unfortunately I think this might just be different in different backends. 

This is what I get for your example:

image

sugatoray:
The outputs in PDF look awe fully good and crisp! I guess vector graphics is the answer to final image quality. 

@srush  Thank you for working on this one!

Issue #38, opened 2022-05-22
Place the connecting arrow between the centers of two diagrams on their corresponding bounding boxes


danoneata:
I don't know if this would improve the visuals of the arrows, but I was wondering whether it would help to emulate the behaviour of [`connectOutside`](https://hackage.haskell.org/package/diagrams-lib-1.4.5.1/docs/Diagrams-TwoD-Arrow.html#v:connectOutside) from `diagrams` (but using bounding boxes instead of traces):

> The arrow lies on the line between the centres of the diagrams, but is drawn so that it stops at the boundaries of the diagrams

Ideally, I would like to be able to draw something [like this](https://diagrams.github.io/doc/quickstart.html#a-worked-example)

<img src="https://user-images.githubusercontent.com/819256/169715241-ab20b659-d393-4a82-8c60-45d2022b1ab8.png" height=200></img>

but without manually specifying the "cardinal" points of each arrow.

@srush What do you think? Is it worth investing time in something like this?

srush:
It is a nice idea, and probably would not be so hard to do. However as you note, ours will not be as pretty since it will be to the square Box and not the "envelope" like in diagrams. 

I agree the Cardinal direction thing is a bit hacky and should be removed eventually if we figure out a better way. (Either way it should probably be tlrb not nwse to match the library). 

We might want to have a file of comp geometry functions if more things like this accumulate.

srush:
Great. I'm not changing anything right now. I have a couple remaining wishlist items, I'll add them to issues. But generally finding it really useful.

srush:
Think we can call this done-ish. Who knew arrows are so complicated.

PR #72, opened 2022-06-18
Implement traces to allow for more flexible connections

danoneata:
This pull request is still work in progress, but I wanted to share it so that I get an opinion that I'm heading in the right direction.

The PR is related to the discussion in #38 and implements [traces](https://hackage.haskell.org/package/diagrams-core-1.5.0/docs/Diagrams-Core-Trace.html#t:Trace), which allow us to place connecting arrows on the perimeter of a diagram. [For example](https://github.com/danoneata/chalk/blob/danoneata/trace/examples/tournament-network.py), we are now able to draw something like this:

image

Currently, I have implemented traces for diagrams made out of paths (rectangles, regular polygons) and circles; for the rest of the diagrams, the trace defaults to the corresponding bounding box.

A remaining challenge is to make the traces work on named subdiagrams: currently, we can only retrieve [the bounding box of a subdiagram](https://github.com/danoneata/chalk/blob/danoneata/trace/chalk/core.py#L536), but ideally we want to access the (sub)_diagram_, such that we get its trace. I can take a stab at this, but I would like an opinion prior to doing that, since I expect to involve some sizeable changes.

Finally, I am aware that the geometric functionally of the library is poorly organized and likely confusing (we now have lines, segments, paths, trails, points, vectors); any advice on how to clean this up is appreciated.


srush:
This is really nice implementation. It's pretty clean as written and seems to preserve all the existing functionality.

My rec would be to commit this functionality, and then we should consider next steps.

For example, it seems like we should remove bounding boxes entirely as a first class object visible to the user? Align is now really trivial right (trace with unit vectors), and besides / atop as well. 

That being said, there are some non-trivial cases (pad?) Also I like viewing bounding boxes, but we can add a `show_trace` as well which samples the trace around a diagram. A couple of the examples also use box arithmetic (tree for example).

srush:
Oh I guess padding is not so bad, it seems like Diagrams does this by padded Envelopes https://hackage.haskell.org/package/diagrams-lib-1.4.5.1/docs/Diagrams-Envelope.html#t:Enveloped (traces from the origin.) And you can get pad_l by composing with a projection.

This seems like the part of the library where the functional parts really shine.

danoneata:
Hello! Sorry for the late reply!

> For example, it seems like we should remove bounding boxes entirely as a first class object visible to the user?

I don't think that bounding boxes are really exposed to the user, are they?

The main issue that I see with bounding boxes is that they don't compose well: we end up with loose bounding boxes after certain transformation. Let's consider the following diagram:
```bash
square(1).rotate_by(1 / 8).show_bounding_box()
```
o1

If we were to add a small circle and rotate back with end up with a looser bounding box:
```bash
(square(1).rotate_by(1 / 8) + circle(0.05)).rotate_by(1 / 8)
```
o2

Note also this doesn't happen if the transformations are applied in sequence (because we first compose the transformations, prior to applying them to the bounding box):
```bash
square(1).rotate_by(1 / 8).rotate_by(1 / 8)
```
o3

> Align is now really trivial right (trace with unit vectors), and besides / atop as well.

Yes, if I understand correctly this is also [how `diagrams` implements the juxtaposition of two diagrams](https://hackage.haskell.org/package/diagrams-core-1.5.0/docs/src/Diagrams.Core.Juxtapose.html#juxtaposeDefault).

But I wanted to draw attention regarding an aspect. They distinguish between two primitives: [envelopes](https://hackage.haskell.org/package/diagrams-lib-1.4.5.1/docs/Diagrams-Envelope.html#t:Enveloped) (which roughly correspond to convex hulls) and [traces](https://hackage.haskell.org/package/diagrams-core-1.5.0/docs/Diagrams-Core-Trace.html) (which roughly correspond to the boundary of the shape). I would assume these two behave differently for non-convex shapes, but I'm not sure if there other implications as well.

srush:
Interesting, I'll have to read some more. The definition of envelope looked superficially to me like a trace from the origin. I wonder if the convexity is enforced by initializing it as the convex hull for paths, and then only applying affine transformations. 

One issue I see is that we currently allow shapes that do not intersect the origin. Think this would be not allowed.

In terms of exposing bounding boxes, we currently has "show_bounding_box" and "get_subdiagram_bounding_box" in the API. But we should remove them for the sake of interestingness.

Finally it does seem like we will implement a lot of 2d geometry. Wonder if it is worth bringing in a dependency for this? Scikit-geom is one possibility, but maybe overkill.

srush:
Actually a bit worried that envelopes might make cat extremely slow in Python, as the number of function calls for boundaries would grow with number of elements. Might be unwieldy without tail calls. 

If we implement them, might consider having it be optional, or consider just allowing polygon boxes.

danoneata:
> (Edit: realize that the need for union makes these pretty hard.)

As far as I understand envelopes are implemented as functions from a direction vector to a distance. The union of two envelopes is just the maximum of the two envelopes:
```python
class Envelope:
    ...
    def __add__(self: "Envelope", other: "Envelope") -> "Envelope":
        return Envelope(lambda d: max(self(d), other(d)))
```

This is also illustrated in Figure 5 from the [Monoids paper](http://ozark.hendrix.edu/~yorgey/pub/monoid-pearl.pdf):
</img>

I think you are correct that we define `Envelope`s of basic diagrams and then the transformations will preserve them, that is:
```
envelope(transform(diagram)) = transform(envelope(diagram))
```

> One issue I see is that we currently allow shapes that do not intersect the origin. Think this would be not allowed.

I'll need to think about this. I guess you are making this observation in the context of `Envelope`s, but I'm not 100% sure that this is an issue.

> In terms of exposing bounding boxes, we currently has "show_bounding_box" and "get_subdiagram_bounding_box" in the API. But we should remove them for the sake of interestingness.

I would be inclined
- to keep `show_bounding_box` since it is a useful utility and it doesn't return a `BoundingBox`, but a `Diagram`;
- prefix `get_subdiagram_bounding_box` with an underscore to denote that it is a private method.

 > Finally it does seem like we will implement a lot of 2d geometry. Wonder if it is worth bringing in a dependency for this? Scikit-geom is one possibility, but maybe overkill.

Thank you for the suggestion! I'll check it out. I don't mind using an existing library, as long as it replaces most of the functionality that we have coded by hand.

> Actually a bit worried that envelopes might make cat extremely slow in Python, as the number of function calls for boundaries would grow with number of elements. Might be unwieldy without tail calls.

I think we can cache the envelopes similarly to what we do for bounding boxes and to what is presented in Variation III of the [monoids paper](http://ozark.hendrix.edu/~yorgey/pub/monoid-pearl.pdf). Would that be enough?

danoneata:
One quick clarification regarding what I've said previously:

> I would assume these two behave differently for non-convex shapes, but I'm not sure if there other implications as well.

In fact, the following example shows that the trace and envelope don't agree even for convex shapes (the green lines go outside the boundary):

</img>

srush:
Ah, that's very helpful. So to summarize.

1) Traces are an independent question from envelopes. We should probably just merge this PR, roughly as is.
 
2) We should remove Boxes from the public API. Particularly subdiagrams. For now we can replace connect with this implementation.

3) Envelopes are interesting but separate than traces. They solve the problems you showed above (and would also solve the tensor.py example in an elegant way, I think, since the cube would be enveloped).

4) Consider a lightweight geometry implementation to reduce library size.

5) I will read the papers to catch up with you 🤣

danoneata:
Good summary! I guess we can merge then. One small question: I haven't had the chance to rename the cardinals from north, south, etc. to top, bottom, etc. as you suggested in the other issue (#38). Should I still do it, or maybe it isn't that critical now that we won't expose the `BoundingBox` to the user?

srush:
I think an elegant way to handle these is by exposing unit vectors for directions? So N is unit_y  / NE is unit_x + unit_y? I think that plays well with traces? The paper makes a convincing cases that bounding boxes / envelopes should not in the end be used for connections.

PR #75, opened 2022-06-27
Replace low-level geometry 

srush:
This PR attempts to simplify and eventually speed up the library by removing the low-level geometry primitives in favor of a third-party library. This also sets things up to make the transition to envelopes a bit easier (such as convex hull functionality). 

The main goal of this PR is described in https://github.com/danoneata/chalk/issues/74 .

* Removes BoundingBox, Point, Vector, Line, Transform classes entirely.
* Removes Affine and remaining non-render calls to cairo.Matrix
* Introduces Optional BoundingBoxes needed for envelopes
* Farms out work for Paths to batch version of transforms. 

Took a bit more time than expected because the diagram code has gotten a bit complex (mostly my fault). Hopefully this simplifies things a bit. 

Gotchas:

* Seemingly there is a bug in the planar library. When you call affine * shape it does the matrix multiple wrong! I have a hack called `_fix_affine` that needs to be called before applying an affine to an object (composing affines works fine). 
* We can no longer override apply_transform on the core geometric objects. This means that things like canceling out translations on vectors no longer works. Be careful. 
* The object we were calling Line is called Ray in planar so switching to that name.


danoneata:
Thanks! At a first glance, looks good—lots of code we can remove! I'll have a more careful look tomorrow.

> We can no longer override apply_transform on the core geometric objects. This means that things like canceling out translations on vectors no longer works. Be careful.

Yes, not having the vectors translation-invariant is a bit unfortunate. I guess it's somehow related to the fact that `planar` doesn't distinguish between points and vectors (since `Point` is defined as `Vec2`); so we also have to be careful not to add points, for example.

chalk/core.py
```diff
@@ -189,35 +198,49 @@ def _repr_svg_(self) -> str:
         os.unlink(f.name)
         return svg
 
-    def atop(self, other: "Diagram") -> "Diagram":
+    def _empty_check(
```
danoneata, line 201: I don't know if this function will be replaced when moving to envelopes, but I find its logic a bit hard to follow; maybe a short comment would be helpful. Some thoughts:

- I assume that it is needed because `Empty` diagrams do not have a `BoundingBox`, right? It wouldn't work if `Empty` diagrams were to return a fake bounding box (`BoundingBox([Point(0, 0)])`) instead of `None`, would it?
- Are the `fake` bounding boxes ever used? If I understand correctly, if `box1 is None or box2 is None` then the first argument returned is not `None` (it is either `self` or `other`) and the caller function (e.g., `_merge` or `atop`) will return it (while the second and third arguments will be ignored).
- Would it be a bit clearer if the conditions were on diagram instead of the bounding box; that is, `if isinstance(self, Empty)` instead of `if self.get_bounding_box() is None`?

chalk/core.py
```diff
@@ -189,35 +198,49 @@ def _repr_svg_(self) -> str:
...
 
     __add__ = atop
 
-    def beside(self, other: "Diagram") -> "Diagram":
-        box1 = self.get_bounding_box()
-        box2 = other.get_bounding_box()
-        dx = box1.right - box2.left
-        t = tx.Translate(dx, 0)
-        new_box = box1.union(box2.apply_transform(t))
+    def _merge(self, other: Diagram, direction: Vec2) -> Diagram:
```
danoneata, line 222: A small observation. It might have been nice for this function to work on arbitrary `direction` vectors; for example, I would have expected the following snippet
```python
square(1)._merge(circle(0.5), 2 * unit_x + unit_y)
```
to place the bounding boxes of the two diagrams touching along the `2 * unit_x + unit_y` axis.

But again, I guess this functionality is already present in the envelopes branch.

chalk/trail.py
```diff
@@ -1,12 +1,13 @@
...
 
-from chalk import transform as tx
 from chalk.core import Primitive
-from chalk.point import ORIGIN, Point, Vector
 from chalk.shape import Path
+from chalk.transform import Transformable, apply_affine
 
 
-class Trail(tx.Transformable):
+class Trail(Transformable):
```
danoneata, line 10: Should we make `Trail` a `dataclass`?

chalk/trail.py
```diff
@@ -15,10 +16,12 @@ class Trail(tx.Transformable):
     with illustrations/figures).
     """
 
-    def __init__(self, offsets: List[Vector]):
-        self.offsets = offsets
+    offsets: Vec2Array
+
+    def __init__(self, offsets: Vec2Array):
+        self.offsets = Vec2Array(offsets)
```
danoneata, line 22: If `offsets` are already of type `Vec2Array` then there is no need to wrap them again in `Vec2Array`.

chalk/trail.py
```diff
@@ -90,8 +83,8 @@ def apply_transform(self, t: tx.Transform) -> "Trail":  # type: ignore
         Returns:
             Trail: A trail object.
         """
-        return self.transform(t)
+        return Trail(apply_affine(t, self.offsets))
```
danoneata, line 86: Should `Trail` be translation-invariant, given that it is a list of vectors?

examples/hex_variation.py
```diff
@@ -42,7 +41,8 @@ def hex_variation(num_tiles):
     return place_at(diagrams, grid)
 
 
-dia = hex_variation(12).line_width(0.02).rotate_by(1 / 4)
+dia = hex_variation(12).line_width(0.05).show_bounding_box()
```
danoneata, line 44: Maybe we can remove the bounding box? But I think I understand why you looked at it: the output is not centered for whatever reason.

chalk/segment.py
```diff
@@ -97,9 +93,11 @@ def line_circle_intersection(line: Line, circle_radius: float) -> List[float]:
     This is a quadratic equation, whose solutions are well known.
 
     """
-    a = line.v.dx**2 + line.v.dy**2
-    b = 2 * (line.p.x * line.v.dx + line.p.y * line.v.dy)
-    c = line.p.x**2 + line.p.y**2 - circle_radius**2
+    p = ray.anchor
+
+    a = ray.direction.length2
```
danoneata, line 98: Much better! 🙂

chalk/core.py
srush: Yeah this is a little awkward. It is based on the Yorgey paper that defines envelopes as Options 

image

Why don't I push this into the Envelope class and its mappend definition. I will fix it in the next PR?

chalk/core.py
srush: Yup that makes sense. I will define this by adding an optional argument to `atop`? 

`square(1).atop(circle(0.5),  2 * unit_x + unit_y)`

Thing that fits with the `diagrams` version.

PR #76, opened 2022-06-28
Envelopes

srush:
PR to add an envelope type to replace bounding boxes. Based on the Monoid Pearl, diagram implemention, and following the format used by traces. 

Little demo: 

image

Envelopes seem to work reasonably well. But they are very counter intuitive! For instance when you show them this is what they look like (at first I thought this was a bug): 

image

Other downsides:

* For now removing pad_l, pad_r etc and changing name of pad to frame. They become a bit strange. 
* Positioning is still a bit non-intuitive to me. Not sure how to explain to users how it works.
* Probably need some richer atop / align functions for this to be actually useful.


danoneata:
Nice work!

> For instance when you show them this is what they look like (at first I thought this was a bug)

Ah, wow, that envelope looks very unexpected indeed at first glance (although now I can see why it is correct)! Maybe "convex hull" (as suggested in the paper) is not the best analogy for envelopes (although maybe we can recover the points on the convex hull by picking various origins)?

But, on the positive side, I think that the most common positioning functions (vertical positioning with `above` and horizontal positioning with `beside`) still behave as expected (and as when we were using bounding boxes).

Also, the envelopes compose well with transformations 🙂 [whereas the bounding boxes didn't](https://github.com/danoneata/chalk/pull/72#issuecomment-1163504560).

```python
square(1).rotate_by(1 / 8).show_envelope()
```
o1

```python
(square(1).rotate_by(1 / 8) + circle(0.1)).rotate_by(1 / 8).show_envelope()
```
o2

srush:
Simplified the alignment and atop logic. Added in a an "is_empty" field for tracking empty envelopes. figures look reasonably good now.

danoneata:
This is very nice! Lots of nifty features (like `juxtapose` and `snug`)! I also like it better that `beside` takes the additional `direction` rather than `atop` :+1: I'll go through the code and leave some comments, but otherwise I think I'm happy with the changes!

One (more general) question: for connections (_e.g._, `connect_outer`), wouldn't it be better to use the traces (rather than envelopes) to place the connecting path? Anyways, I think we can finish this pull request first and think about that afterwards

chalk/__init__.py
```diff
@@ -216,11 +217,12 @@ def beside(diagram1: Diagram, diagram2: Diagram) -> Diagram:
     Args:
         diagram1 (Diagram): Left diagram object.
         diagram2 (Diagram): Right diagram object.
+        direction (Optional[Vec2]): Placement direction.
```
danoneata, line 220: The documented type (`Optional[Vec2]`) doesn't agree with the given type (`Vec2`).

chalk/core.py
```diff
@@ -198,47 +204,44 @@ def _repr_svg_(self) -> str:
...
-
     def above(self, other: Diagram) -> Diagram:
-        return self._merge(other, unit_y)
+        return self.beside(other, unit_y)
 
     __truediv__ = above
 
-    def beside(self, other: Diagram) -> Diagram:
-        return self._merge(other, unit_x)
+    def beside(self, other: Diagram, direction: Optional[Vec2]) -> Diagram:
```
danoneata, line 240: From my reading, I think that the type of `direction` should be `Vec2`, as `juxtapose` doesn't know what to do with a `None`.

danoneata:
And one final question: do you have any idea why for the `logo` example, the output's aspect ratio seems off? If not, I can debug that myself.

srush:
The logo issue is my fault. Since I removed pad_l we don't have a way to reduce sizing in one direction. I will send a next PR to clean that up (I think you need to have a padding that projects along a dimension, but I didn't know how to do it). I also need to update the docs.

PR #77, opened 2022-06-29
Sync master with envelopes

srush:


danoneata:
Thanks @srush ! I'll fix the documentation over the weekend!

srush:
Cool, actually have a PR coming that fixes the docs. Sorry about that!

danoneata:
Ah, sure, no worries! Thanks for letting me know! I can look into other things (connections or subdiagrams).

srush:
Yeah we should think about what to add next. Having too much fun playing with this. Better Subdiagrams sounds great. I also think we could simplify the Diagram class hierarchy a bit. Think there is some overlapping functionality.

 I'm going to go back and just fix some small things, and work on documentation. Would be fun to get some users.

My short-term todos:
* Hide planar classes from the user
* Make show_envelope easier to read
* Write some more docs.
* add a couple more envelope functions

Issue #35, opened 2022-05-22
Splines


srush:
Lo-pri: The Haskell cubic spline api seems pretty doable now that we have paths / trails. I think this could be implemented as a Spline shape that takes a path and renders it using https://docs.scipy.org/doc/scipy/tutorial/interpolate.html

https://hackage.haskell.org/package/diagrams-lib-1.4.5.1/docs/Diagrams-CubicSpline.html#g:1

danoneata:
Yes; if we want curvy segments, we could also have Bézier curves, which would be simpler to render since they are already supported by both PyCairo and SVG. But I assume that cubic splines are somewhat more intuitive to specify?! Out of curiosity, are there some diagrams that you would like to draw and would require splines?

srush:
Yeah it looks like diagrams converts splines to bezier curves internally. Guess splines are easier for users.

No I don't actually need splines. Bends are enough for me. Just seemed interesting.

PR #78, opened 2022-07-03
Tutorial: Big Ben

srush:
Kind of an extended tutorial demo to replicate a diagram of the Big Ben clockface. Would love thoughts or comments for simplifying further.

https://srush.github.io/pydiagrams/examples/bigben/

image

Here's what it really looks like:

image

This is not ready to check in yet, as there are some major breaking changes. These include: 

* Switching everything user facing to degrees from radians. 
* Changing the way the show_* functions work.
* Monkey patching geometry so it works in the OO style. 
* Fixing some bugs with arcs and circles. 
* Making line_width not scale in SVG (vector-effect style)

Sorry these probably should be different PRs, but most of them came up in the process of making the demo.


chalk/shape.py
```diff
@@ -242,28 +243,44 @@ class Arc(Shape):
...
         self.ctx = cairo.Context(surface)
 
     def get_bounding_box(self) -> BoundingBox:
         self.render(self.ctx)
         l, t, r, b = self.ctx.path_extents()
-        return BoundingBox([Point(l, t), Point(r, b)])
+        return BoundingBox([P2(l, t), P2(r, b)])
+
 
+    def get_envelope(self) -> Envelope:
```
danoneata, line 256: I've noticed in the Big Ben tutorial that the `under_arc` curve is chopped off when rendered. On closer inspection, it seems that there is a slight issue with the envelope. An arc can, in principle, be specified by negative angles:
```python
 arc(radius=1, angle0=-50, angle1=50)
```
but in the `wrapped` function the direction `d` takes only positive angles; so if a direction's angle is in the `[310, ..., 360]` range then the `if` condition enters through the second branch ("point outside of arc"). For the above example, the envelope looks like this (see the upper half of the envelope, which corresponds to the `[-50, ..., 0]` range):
o

srush:
Okay, think this roughly works now, however I realize there are some big questions that I may have broken. 

* I made line-thickness scale invariant for SVG. As a user, I think you really need this for scaling compositionality to work. But it is counterintuitive that you can't really scale your diagrams after making them. Other backends probably need this as well.  Maybe it should be optional?

* There are some inconsistencies with angles. I can't decide if they should go clockwise or counterclockwise, i.e. this seems bad
image

srush:
Oh, as always the Diagrams manual answers my questions. 

* Line width-scaling should be in output normalized units  by default https://diagrams.github.io/doc/manual.html#measurement-units with optional local units. 

* Counter clockwise is correct. I should override or hide the .angle property from users.

danoneata:
I've made a small change to the computation of the arc's envelope. I hope didn't break anything and it works fine now—is this the behaviour you would expect?
```python
arc(radius=1, angle0=-45, angle1=90)
```
o
```python
arc(radius=1, angle0=-45, angle1=360)
```
o
> There are some inconsistencies with angles. I can't decide if they should go clockwise or counterclockwise, i.e. this seems bad

Yes, I have to fix this for arcs as well: currently, negative angles are in the top half and the arc is drawn clockwise...

If we go with the counterclockwise convention, maybe we should have `unit_y = Vec2(0, -1)`, such that, `unit_x.rotate(90) == unit_y`?

srush:
Thanks. this is great, that was bugging me. (I am pushing a couple other tweaks I made as well, realize the colors are off in your version).

Another related thing is that it would be nice if `Path` and maybe `Trail` could have arcs in it. Wanted to have the hands use an arc_between, but couldn't fill.

danoneata:
> Another related thing is that it would be nice if `Path` and maybe `Trail` could have arcs in it. Wanted to have the hands use an `arc_between`, but couldn't fill.

Yes, that's a good point! I've added an issue as a reminder. It might be also a good opportunity to allow for more flexible curves, such as Bézier curves or splines (as you suggested in #35).

srush:
Gonna merge this in. Seems like it now fixes more than it breaks. Made issues for the remaining problems.

Issue #69, opened 2022-06-16
Better Text bounding boxes


srush:
For text bounding boxes we are reliant on calling PyCairo TextExtents. As far as I can tell this function call is just broken, or something else is going wrong:

![image](https://user-images.githubusercontent.com/35882/174086638-5aaeaab2-05e7-4f11-a3f4-437fa51499f2.png)

One solution is that if we know the exact font there are ways of calculating sizes. However this might rely on having a font distributed with the library.

danoneata:
Yup! I agree!

Issue #68, opened 2022-06-16
Make PyCairo Dependency Optional


srush:
The Cairo API is really nice, but it is a pain to install in many contexts. When working in a notebook, the SVG backend seems fine and has minimal dependencies. 

Proposal: 

* Replace the Transformations class with Affine: https://github.com/rasterio/affine
* Remove all the SVG renderers in transformations. 
* Write converters from Affine to SVG "matrix" and cairo "Matrix"
* Figure out a better solution for SVG text sizes than calling Cairo.

srush:
`!pip install https://github.com/danoneata/chalk/` now works without dependencies.

PR #81, opened 2022-07-07
Remove Cairo prerendering dependencies

srush:
Should clear up these two issues. 

https://github.com/danoneata/chalk/issues/69
https://github.com/danoneata/chalk/issues/68

In theory now you can run the SVG backend without any non-python dependencies. Think that fits a lot of notebook like use cases.


danoneata:
Thanks! Looks good!

Just a couple of minor points that are not strictly related to this PR:
- Regarding the logo's aspect ratio, what about using the `with_envelope` to do the cropping? Something along these lines:
```diff
diff --git a/examples/logo.py b/examples/logo.py
index c350660..b6698d1 100644
--- a/examples/logo.py
+++ b/examples/logo.py
@@ -43,12 +43,12 @@ def sunflower(n):
 floret = sunflower(1900).center_xy().scale_uniform_to_x(1).center_xy()
 background = rectangle(1.5, 1).fill_color(black).line_width(0).translate(-0.25, 0)
 logo = text("Chalk", 0.35).fill_color(grey).line_width(0.008).line_color(black).translate(-0.15, -0.1)
-mask = rectangle(1.5, 1).fill_color(white).line_color(white).line_width(0.05).translate(0, 0.6)
+mask = rectangle(1.5, 0.6)

 # assemble
-d = (background + floret + logo).center_xy() + mask
+d = (background + floret + logo).center_xy().align_t().with_envelope(mask.align_t())
```
- I don't know, why but I get different outputs for the examples than those newly added; in particular, I see that
  - the added `tournament-network.svg` doesn't have the digits better centered in the circle nodes
  - the added `logo.svg` seems to have the text cut.
- I've noticed that the `mypy` tests have failed; I guess we need to `type: ignore` those lines!?

srush:
Good eyes. 

* Fixed logo 
* Fix for tournament image
* Forced MyPy newer version.

Issue #80, opened 2022-07-07
Sync line widths across backends


srush:
Following `diagrams` plan is to use "normalized" units for setting line_widths and arrow sizes. 

The way this works is that there will be two commands `line_width_local` and `line_width`. The first will scale with the transformations applied to it, and the second will be a scale_invariant, but will be a fraction of the final document size. 

This is implemented for the SVG backend using the `vector-effect` command. For PNG and PDF need to think through the best way to do it.

PR #82, opened 2022-07-08
Unify line width across backends

srush:
This turned out not to be so bad. Assumes width is proportional to the largest dimension of the output. Doesn't let you "scale" width at all. 

PDF -> This is the default behavior
PNG -> moved line_width out of linear transformation
SVG -> uses a vector-effect command. 

https://github.com/danoneata/chalk/issues/80

For now there is a "line_width_local" option, but I don't have a good use case for it yet, so it is hidden. It seems like for most of the examples normalized is the right measure. 

Also added a little webpage for visually checking that all the pictures look the same. Kind of a hacky version of a unit test. 

(Note one weird bug of the SVG sizing is that they have to be shown at the right size. If you resize them after rendering they look bad. Not sure how to fix this yet)


danoneata:
Nice! Looks good to my eye! Only a few small remarks:

- I feel that I get more consistent widths across the three backends with the following normalization constants:
```python
# for cairo's PNG
normalizer = self.output_size * (15 / 500)
# for SVG (the same one that you set)
normalizer = self.output_size * (15 / 500)
# for PDF
normalizer = self.output_size * (175 / 500)
```
- The line widths for some of the examples could be tweaked a bit: in particular, the `koch`, `sqaures`, `tournament-network` examples could use a thicker line width, while `escher-square-limit` a bit thinner line? I'm fine doing these updates myself.

> Also added a little webpage for visually checking that all the pictures look the same. Kind of a hacky version of a unit test.

Thanks! A few questions and observations on this as well:
- What is the command to run to obtain the HTML output?
- I think there isn't any `intro-00` example (although it appears in the `examples/tests/output.html`); and maybe we should add the `normalized` example's output to the repo as well?
- What about using [`object`](https://stackoverflow.com/a/4482962/474311) instead of `img` for embedding the SVGs in the HTML? It looks that the lines are thinner when using `img`?

srush:
Great. Made all of these changes except for updating the examples. 

Added a command `make serve` that serves this page on 8080. The file is now in `examples/output/index.html`

The command regenerating the index file is: 

`jinja2 examples/tests/index.tpl.html > examples/output/index.html`

But we shouldn't needed it too much.

Issue #84, opened 2022-07-17
`Planar` install issue. 


srush:
It seems like planar relies on a conversion library 'build_py_2to3' that is deprecated in the latest versions of distutils. This is a note to myself to fix this. 

I think the best way to do this may be to just host a modern version of this code. My fault for using an old library.

@renhaofan

I try to run command pip install git+https://github.com/danoneata/chalk@srush-patch-1, But an error occurred:

 × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [13 lines of output]
      Traceback (most recent call last):
        File "/tmp/pip-install-ofq9w8yg/planar_3384c07b07cb49e694e3d9f59655d633/setup.py", line 11, in <module>
          from distutils.command.build_py import build_py_2to3 as build_py
      ImportError: cannot import name 'build_py_2to3' from 'distutils.command.build_py' (/home/steve/.conda/envs/py37/lib/python3.7/site-packages/setuptools/_distutils/command/build_py.py)
      
      During handling of the above exception, another exception occurred:
      
      Traceback (most recent call last):
        File "<string>", line 36, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-ofq9w8yg/planar_3384c07b07cb49e694e3d9f59655d633/setup.py", line 14, in <module>
          raise ImportError("build_py_2to3 not found in distutils - it is required for Python 3.x")
      ImportError: build_py_2to3 not found in distutils - it is required for Python 3.x
      [end of output]
And my python version is totally the same with Colab.

>>> sys.version
'3.7.13 (default, Mar 29 2022, 02:18:16) \n[GCC 7.5.0]'

pawamoy:
Sorry for necro-bumping, but I experience this very issue today still, see https://github.com/chalk-diagrams/planar/issues/1. I'm not sure to understand if this was fixed but not released, or not fixed at all.

PR #85, opened 2022-07-17
Planar replace

srush:
People are reporting that Planar is not working on new versions of python. This is due to the C extension which is slow to install, and which we are note using anyway. This PR replaces the pip planar with a hosted one that just keeps the python code and removes the outdated element. 

This is not a good long term solution, but will come up with a better one later.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -19,7 +19,6 @@ jobs:
         sudo apt-get install libcairo2-dev pdf2svg texlive texlive-science texlive-latex-recommended texlive-latex-extra
         python -m pip install --upgrade pip
         pip install -U pip wheel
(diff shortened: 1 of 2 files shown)


danoneata:
Thanks! Does this pull request fix #84? I'm not sure I see the difference between this branch (`planar-replace`) and the other one (`srush-patch-1`), which seems to fail in #84.

srush:
Yeah, I think it fixes https://github.com/danoneata/chalk/issues/84 . I pulled this into `srush-patch-1`, not sure why it isn't working for that user. 

`srush-patch-1` was a small patch to try to hack in `Output Units` https://diagrams.github.io/doc/manual.html#output-units I'll do it for real later.

Issue #89, opened 2022-07-23
Split up `core` into multiple files.


srush:
The file core.py is getting really big and seems to do too many things. Would like to split up core.py to look more like the diagrams hierarchy (align.py, render.py, juxtopose.py, arrow.py etc.) However the OO style is convenient for the user. 

There isn't a nice way to do this in Python, but this seems not so bad.

https://stackoverflow.com/questions/47561840/how-can-i-separate-the-functions-of-a-class-into-multiple-files

One side benefit of this is that every function could in theiry be used in either functional or OO calling convention. (Like atop now).

danoneata:
Hello! That sounds good! I'm in favour of splitting the `core.py` module in submodules based on their functionality and using a more functional style.

A question that I have is whether we are going to end up with some circular imports and how to fix this. For example, let's say we have a `core/render/svg.py` submodule that defines the `to_svg` and `render_svg` functions. I would expect this module to need importing the data types (for type hinting and maybe also for dispatching based on the data type):
```python
from chalk.types import Diagram, Primitive, Empty

def to_svg(diagram: Diagram, dwg: Drawing, style: Style) -> BaseElement:
    if isinstance(diagram, Primitive):
        ...
    elif isinstance(diagram, Empty):
        ...
    ...

def render_svg(diagram: Diagram, **kwargs):
    ...
    # uses to_svg
```
Conversely, in `core/types.py` I would expect the `render_svg` function to be imported in order to attach it to the `Diagram` object. These mutual imports will generate an `ImportError`; but maybe there is an easy fix to this.

The functions that traverse the `Diagram` abstract syntax tree (like `to_svg` or `to_tikz`), will need to pattern match on the data type when written in a more functional style. There are probably multiple ways of achieving this: `isinstance` checks (as above), [structural pattern matching](https://peps.python.org/pep-0635/) (available only from Python 3.10); [the visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern#Python_example). I'm curious which approach you would prefer.

srush:
Yes, this is a very valid concern. I do not know what the best answer is, and need to do some more research. 

I was hoping to circumvent this problem for now by not factoring rendering. By starting with only functions of the base class of `Diagram`, I think we could get away with just calling other functions on diagram that we import.

PR #90, opened 2022-07-25
Arrows, connection, and closed paths

srush:
This PR makes an attempt to add Arrows following mostly the Diagrams API. https://diagrams.github.io/doc/arrow.html

I tried to mimic Diagrams here as it is complex. It mostly builds from a simple arrow primitive to more complex structures like connection. 

image

image

(same code in haskell. their code is a bit cleaner because they use lenses for `ArrowOpts`, but I think dataclasses is fine. Maybe we can make a `Stylable` class for creating Style dataclasses incrementally.)

```haskell
oct  = octagon 1 # lc darkgray # lw ultraThick # showOrigin
dias = oct # named "first" ||| strut 3 ||| oct # named "second"

-- Connect two diagrams and two points on their trails.
ex12 = dias # connect' (with & lengths .~ veryLarge) "first" "second"
            # connectPerim' (with & lengths .~ veryLarge)
       "first" "second" (15/16 @@ turn) (9/16 @@ turn)

-- Place an arrow at (0,0) the size and direction of (0,1).
ex3 = arrowAt origin unit_Y

example2 = (ex12 <> ex3) # centerXY # pad 1.1
```

Changes:

* Use traces to implement connect_outer and connect_perim
* Draws arrow heads with the diagram framework, no relying on backend
* Allows trails for jagged arrows and arc_between for curved arrows (we can replace this when trails allows curves).
* ArrowOpts class mimics the datastructure from Diagram for arrow options. 
* Replaces older connect and connect_outer api.
* ArrowHead is specialized to only draw after we know the size. This will become something like "scale-invariant."
* Add a `get_subdiagram_trace` for the tournament diagram. 

* Also changes the renderers to close paths if the last and first are the same position. This should be handled better in the future. 

Issues:

* Need to better understand how arrow-heads should be sized. 
* It's all in Core, because I wanted it to be on the Diagram class. This needs to be fixed see - > https://github.com/danoneata/chalk/issues/89
* The type system is still not great. Will fix after we pull things into individual files.


danoneata:
Thanks! That's awesome! At a first glance, it looks really tidy! I'll take a closer look at the code tomorrow. Just a few comments:

- It seems that the directions for `connect_perim` don't agree with those specified by `Trail`s; I've tried to have a look, but couldn't quickly find the source of discrepancy. Here is an example:
```python
c = circle(0.5).fill_color(orange).named("c")
c = (
    c.connect_perim("c", "c", unit_x, unit_y, ArrowOpts(arcHeight=0.5))
    + trail.unit_x.scale(0.5).stroke()
    + trail.unit_y.scale(0.5).stroke()
)
```
o
- This is minor, but for `ArrowOpts`'s attributes would you mind using `snake_case` (instead of `camelCase`), for consistency with the rest of the code?

srush:
Okay, think these are fixed. 

The issue with the camel case seems to be have been that trace_v returned negative traces first. I think it maybe shouldn't return negatives? seems to go against the point. 

image

danoneata:
I think the previous behaviour, of returning the smallest (most negative) distance was consistent with `traceV` from `diagrams`.

image

I've took the liberty and added the `max` variants (`max_trace_v` and `max_trace_p`), which return the points in the direction of the given vector, and I've used those in the `connect` functions.

Also, I've noticed that `connect_outer` used `tr1` with `env2`:
```python
vs = tr1.trace_v(env2.center, -v) + env2.center
```
Was that correct? I've changed that to apply `tr1` to `env1`.

Finally, I think I've stumbled upon a bug caused by the interplay of `planar.Ray`, which automatically scales the direction to unit length, and the `line_segment` intersection function, which assumes that the length of the segment is preserved in the corresponding `Ray`'s `direction`. i'll open a separate issue for this.

danoneata:
> I think that was on purpose. For connect outside we want the point on the trace of t1 as seen from env2 right? Although maybe it is better to replicate your behavior of checking both from a middle point. When are they different?

Ah, okay, sorry! I've reverted to the previous way of computing the points on the trace:
```diff
-        ps = tr1.max_trace_p(env1.center, v)
-        pe = tr2.max_trace_p(env2.center, -v)
+        ps = tr1.trace_p(env2.center, -v)
+        pe = tr2.trace_p(env1.center, v)
```
But I think they might be equivalent; at least, I haven't observed a difference between the two on our examples.

I think the version that used the midpoint was based on diagrams, see [`connectOutside'`](https://hackage.haskell.org/package/diagrams-lib-1.4.5.1/docs/src/Diagrams.TwoD.Arrow.html#connectOutside%27). I'm not sure I see what are the trade-offs between our current version the midpoint one.

srush:
Gotcha. Let's do midpoint then. As always, I'm sure there is some case with non-centered, non-convex diagrams where the simple solution fails.

danoneata:
Okay, I've made the following changes:
- reverted to using `midpoint` in `connect_outside`
- fixed the issue described in #91 
- made some other small changes for consistency.

I'll let you have a final look on these changes, but otherwise I'm happy with the PR (thanks again!), so feel free to merge at your will.

PR #96, opened 2022-08-05
[WIP] Rewrite of Path to use Segments

srush:
Add a new segment type to path that allows for elliptical arcs and straight line segments. 

Currently this just does Path, but maybe should do trail as well (or we should just have trail? segments assume that it is connected, but located).

Experiment with:
* Rewriting arc, circle, rectangle, etc using path
* Verified that these look the same in different backend. (this was much harder than I was expecting due to the fact that svg uses a non standard arc representation )
* Add some more protocols for using transforms and traces. 
* Rounded rect now in standard form. 

image

Questions: 

* Should these renderers be part of the backend now?
* Can we eliminate Shape class entirely?


danoneata:
Thanks Sasha! I'll take a look in the following days!

srush:
This got a bit more ambitious. Tried to mimic the visitor pattern for shapes as well and move that to the backend. Also realized we weren't importing the types from the planar library, so I had to fix a bunch of typing issues to get it to type check. blech.

examples/path.py
```diff
@@ -0,0 +1,56 @@
+from chalk import *
+from chalk.segment import Segment
+from chalk.arc import ArcSegment
```
danoneata, line 3: Should probably be
```python
from chalk.shapes.segment import Segment
from chalk.shapes.arc import ArcSegment
```

chalk/shapes/__init__.py
```diff
@@ -89,11 +89,11 @@ def rectangle(
     Returns:
         Diagrams
     """
-    return Primitive.from_shape(Rectangle(width, height, radius))
+    return Path.rectangle(width, height).stroke()
```
danoneata, line 92: The `radius` argument is not used.

danoneata:
I've started going through the code and I like the changes: i think it improves the structure of the codebase! I still need to try some more examples, but a potential current issue is that the rendering of the `path.py` (the bottom shape) example looks different as an SVG compared to its PNG or PDF rendering. Also the PDF rendering produces some small imperfections; for example, the origin has a tiny extrusion
image
and in the last example from `path.py` the stroke line is not perfectly closed:
image

chalk/shapes/__init__.py
srush: Ah yeah. Maybe I will do this as a second PR. I want to make sure the rounding is the same in each of the different backends.

srush:
Okay, think I got these. The issue seems to be that Tikz is extremely sensitive to boundary points. The recommendation is to close on line segments if possible. 

https://tex.stackexchange.com/questions/350015/how-can-i-close-a-path-made-of-arcs

image

For circles I made it 2 arcs that match at exactly the right angle. 

image

Unrelatedly. Should we make the cairo backend scale by default to a much higher DPI? at the current default it is very hard to see.

chalk/types.py
```diff
@@ -1,35 +1,49 @@
...
+class SegmentLike(Enveloped, Traceable, Transformable):
+    @property
+    def p(self) -> P2:
+        ...
+
+    @property
+    def q(self) -> P2:
+        ...
+
+    dangle: float
```
danoneata, line 41: This is minor, but I wonder whether `dangle` should be a common attribute to `Segment` and `ArcSegment`, given that it's not really relevant for the `Segment` subclass. I assume that you have included it for its use in the `arrow.py` module, but maybe it can be removed if we explicitly check that the segment is an `ArcSegment` prior to rotating (although I understand this is not a particularly elegant solution):
```diff
diff --git a/chalk/arrow.py b/chalk/arrow.py
index ecd0422..7e56984 100644
--- a/chalk/arrow.py
+++ b/chalk/arrow.py
@@ -94,8 +94,8 @@ def arrow(length: float, style: ArrowOpts = ArrowOpts()) -> Diagram:
             P2(0, 0), P2(l_adj, 0), style.arc_height
         )
         shaft = Path([segment]).stroke()
-        φ = segment.dangle
-        arrow = arrow.rotate_rad(φ)
+        if isinstance(segment, ArcSegment):
+            arrow = arrow.rotate_rad(segment.dangle)
```

srush: yeah, that's a better solution.

srush: Actually is there a nice way to do Union types in python now? Maybe segment should just be `Segment | ArcSegment` since I don't think there are going to be more segment types.

chalk/shapes/arc.py
```diff
@@ -0,0 +1,162 @@
...
+            .scale(r)
+            .rotate_rad(φ)
+            .translate(d / 2, dy)
+        )
+
+        ret = ret.rotate(-diff.angle).translate_by(p)
+        return ret
+
+    def reflect_y(self) -> ArcSegment:
+        assert False
```
danoneata, line 159: I've noticed that the `reflect` transformations are not defined. Is there a particular reason for this? Isn't the default behavior (specified by `Transformable`) correct?

chalk/types.py
danoneata: Yes, I was also thinking of using a [Union type](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#union-types), but I was not sure if would break anything.

danoneata:
> Okay, think I got these. The issue seems to be that Tikz is extremely sensitive to boundary points. The recommendation is to close on line segments if possible.

Great!

> Here is the cairo fixed rect.
> 
> image

Nice! 👍 

> Unrelatedly. Should we make the cairo backend scale by default to a much higher DPI? at the current default it is very hard to see.

Yes, I'm happy to output larger diagrams for `cairo` by default (the current height of 128 indeed results in tiny images).

chalk/shapes/arc.py
srush: I got a bit stuck here. The problem is that reflection changes that sign of `dangle` I wasn't sure how to preserve this information in my parameterization. Given that reflection of a segment within a path (as opposed to the full path) seemed rare, I decided not to handle it.

srush:
> I guess that is what you mean by "higher DPI", right?

I guess the problem is that height also controls line thickness now. It would be nice if that was the same for the same size with PNG / SVG. Maybe PNG could have a seperate scale parameter that defaults to 3 or something. Not sure if this function helps.  

image

chalk/types.py
srush: No that sounds right to me. Perhaps `Shapes` should also be a `Union` since they need to be closed class now for rendering. Visitor is perhaps too fancy.

chalk/shapes/arc.py
danoneata: Okay, I think I see what's the problem. Maybe it can be fixed if we take into account the sign of the determinant of the transformation matrix (a negative determinant means that the direction of the arc changes); something along the following lines:
```diff
diff --git a/chalk/backend/svg.py b/chalk/backend/svg.py
index 72f382e..f02edbf 100644
--- a/chalk/backend/svg.py
+++ b/chalk/backend/svg.py
@@ -110,7 +110,7 @@ class ToSVGShape(ShapeVisitor[BaseElement]):
         elif isinstance(seg, ArcSegment):
             "https://www.w3.org/TR/SVG/implnote.html#ArcConversionCenterToEndpoint"
             f_A = 1 if abs(seg.dangle) > 180 else 0
-            f_S = 1 if seg.dangle > 0 else 0
+            f_S = 1 if seg.t.determinant * seg.dangle > 0 else 0
             return f"A {seg.r_x} {seg.r_y} {seg.rot} {f_A} {f_S} {seg.q.x} {seg.q.y}"
         else:
             assert False, "not a known segment"
diff --git a/chalk/backend/tikz.py b/chalk/backend/tikz.py
index 725c5c1..8a8ae8a 100644
--- a/chalk/backend/tikz.py
+++ b/chalk/backend/tikz.py
@@ -110,9 +110,9 @@ class ToTikZShape(ShapeVisitor[PyLatexElement]):
             start = (seg.p - seg.center).angle
             end = (seg.q - seg.center).angle
-            if seg.dangle < 0 and end > start:
+            if seg.t.determinant * seg.dangle < 0 and end > start:
                 end = end - 360
-            if seg.dangle > 0 and end < start:
+            if seg.t.determinant * seg.dangle > 0 and end < start:
                 end = end + 360
             end_ang = end - seg.rot
             pts._arg_list.append(
```
This change seem to produce the desired results on a simple example (I have defined the `stroke` method on `ArcSegment`):
```python
from chalk import *
from chalk.shapes import ArcSegment

dia = cat(
    [
        ArcSegment(180, 135).stroke().show_origin(),
        ArcSegment(180, 135).scale_x(-1).stroke().show_origin(),
        ArcSegment(180, 135).stroke().scale_x(-1).show_origin(),
    ],
    unit_x,
    sep=0.5
)

dia.render("examples/output/test_arc.png")
dia.render_svg("examples/output/test_arc.svg")
dia.render_pdf("examples/output/test_arc.pdf")
```
test_arc

chalk/types.py
danoneata: Yes, you are probably right; I guess an upside of the current (visitor-based) implementation is that the `Shape` rendering implementation is a bit more consistent with the `Diagram` one :slightly_smiling_face: I'm fine with either option.

danoneata:
> I guess the problem is that height also controls line thickness now.

Hmm... does it? Isn't the line width normalized by the output size by default?

I've tried the `set_device_scale` function, but it doesn't seem to be doing what we were expecting...

chalk/shapes/__init__.py
```diff
@@ -37,9 +43,7 @@ def arc(radius: float, angle0: float, angle1: float) -> Diagram:
       Diagram
 
     """
-    return Primitive.from_shape(
-        Arc(radius, to_radians(angle0), to_radians(angle1))
-    )
+    return Path([ArcSegment(angle0, angle1 - angle0).scale(radius)]).stroke()
```
danoneata, line 46: I don't know how important this is, but should the `arc` function to use the same direction (either clockwise or counterclockwise) irrespective of its arguments? For example, 
```python
dia = cat(
    [
        arc(1, 0, 90).show_origin(),
        arc(1, 90, 0).show_origin(),
        arc(1, 90, 360).show_origin(),
    ],
    unit_x,
    sep=0.5
)
```
produces
test_arc
but maybe the second arc should be the same as the third one?

chalk/types.py
srush: Okay, moved Segment to be a union. Left shape as is.

chalk/shapes/__init__.py
srush: I have no idea what is right here. As far as I can tell the different rendering libraries make different choices.

chalk/shapes/arc.py
srush: What a nice solution! Added.

srush:
Think we are converged here. Going to check this in and add issues for: 

* `Arc` arguments
* Rounded rect implementation and example (arrows to corners)
* Changes to PNG.
* Trail / Path. Perhaps most of the logics should be moved to Trail and path should just be a trail and a starting point.

danoneata:
> Think we are converged here.

Agreed! 🙂 Thanks for the PR!

Issue #95, opened 2022-08-04
Monoids


srush:
Given Diagrams is basically a ode to monoids, it is a bit disappointing that we can't type these in Python. Protocols can represent `<>` well, but as far as I tell can't do `mempty` since it is type dependent. 

Alternatively we could use multiple inheritance and have a monoid base class with empty / <> defined. That would at least allow a generic fold.

danoneata:
Interesting! I assume you are suggesting something along these lines?

```python
from __future__ import annotations

from abc import ABCMeta, abstractmethod
from dataclasses import dataclass
from typing import List, TypeVar


A = TypeVar("A", bound="Monoid")


class Monoid(metaclass=ABCMeta):
    @abstractmethod
    def mappend(self: A, other: A) -> A:
        pass

    @staticmethod
    @abstractmethod
    def mempty() -> A:
        pass

    @classmethod
    def mconcat(cls, elems: List[A]) -> A:
        accum: A = cls.mempty()
        for elem in elems:
            accum = accum.mappend(elem)
        return accum


@dataclass
class All(Monoid):
    """Boolean monoid under conjunction."""

    value: bool

    def mappend(self: All, other: All) -> All:
        return All(self.value and other.value)

    @staticmethod
    def mempty() -> All:
        return All(True)


res1: All = All.mconcat([])
res2: All = All.mconcat([All(True), All(False)])

print(res1)
print(res2)
```
Unfortunately, I'm having problems typing even this code:
```
monoid.py: note: In class "All":
monoid.py:39:5: error: Return type "All" of "mempty" incompatible with return type "A" in supertype "Monoid"  [override]
        def mempty() -> All:
```
Do you know if I'm doing something obviously wrong? Or, alternatively, did you manage to come up with a version that type checks?

srush:
For some reason this works. Pretty mysterious. 

```python
from __future__ import annotations                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                            
from abc import ABCMeta, abstractmethod                                                                                                                                                                                                                                                     
from dataclasses import dataclass                                                                                                                                                                                                                                                           
from typing import List, TypeVar, Generic, Type                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                            
A = TypeVar("A", bound="Monoid")                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                            
[...]

PR #102, opened 2022-08-12
Trails replace paths

srush:
This PR replaces the functionality of Paths with a better Trail object. The idea of the trail is that it holds Translation Invariant segments. This is implemented for now by wrapping the segments we have in 0 centered versions. (although we probably can get rid of the Located* versions, these aren't used). All the path drawing is now moved to trails, which compose a bit nicer since they have a well defined monoid. 

Changes:

* `Located` object for trails with a starting point. 
* Segments / Arc segments act like trails as well which makes it more concise, see `TrailLike`
* Paths now can take multiple located trails. This makes it easy to draw things like rings. 
 
image

* Trails have each of the individual shapes so you can compose in Trail space. 
* Trails need to be explicitly closed, otherwise they don't fill. 
 
image

(This is probably my last big core PR, semester starting up again :( I'll try to fix my bugs though.  )


danoneata:
This is awesome! Great job!

I've played a bit with the code and it was a reasonably smooth experience :+1: I've tried to recreate the [Haskell example that allows for multiple interpretations using various monoid instances](https://github.com/danoneata/chalk/issues/14#issue-1231146989):
```haskell
ts = mconcat . iterateN 3 (rotateBy (1/9)) $ triangle 1
example = (ts ||| strokeP ts ||| strokeLine ts ||| fromVertices ts) # fc red
```
As expected it is more verbose and explicity in our case, but I think it does the job:

```python
from colour import Color
from chalk import *
from toolz import iterate, take

def centroid(t):
    points = t.points()[1:]
    return sum(points, V2(0, 0)) / len(points)

red = Color("red")

t = Trail.regular_polygon(3, 1)
t_loc = t.at(-centroid(t))

dia1 = concat(take(3, iterate(lambda d: d.rotate_by(1 / 9), t_loc.stroke()))).fill_color(red)
dia2 = Path.concat(take(3, iterate(lambda d: d.rotate_by(1 / 9), t_loc.to_path()))).stroke().fill_color(red)
dia3 = Trail.concat(take(3, iterate(lambda d: d.rotate_by(1 / 9), t))).stroke().center_xy()

dia_all = hcat([dia1, dia2, dia3], sep=0.2)
dia_all
```
o

I'm going to leave some initial comments and see if I find anything else in the next couple of days.

chalk/shapes/path.py
```diff
@@ -1,122 +1,83 @@
...
 def make_path(segments: List[Tuple[float, float]]) -> Diagram:
     return Path.from_list_of_tuples(segments).stroke()
 
 
-SegmentLike = Union[Segment, ArcSegment]
-
-
 @dataclass
-class Path(Shape, tx.Transformable):
+class Path(Shape, Enveloped, Traceable):
```
danoneata, line 21: Shouldn't the `Path` be still `Transformable`?

chalk/trail.py
```diff
@@ -1,110 +1,166 @@
...
+    def get_trace(self) -> Trace:
+        return Trace.concat(
+            segment.get_trace().translate_by(location)
+            for segment, location in self.located_segments()
+        )
+
+    def stroke(self) -> Diagram:
+        return self.to_path().stroke()
+
+    def apply_transform(self, t: Affine) -> Located:  # type: ignore
```
danoneata, line 56: Shouldn't the code here rather be:
```diff
     def apply_transform(self, t: Affine) -> Located:  # type: ignore
-        t2 = remove_linear(t)
+        t2 = remove_translation(t)
         return Located(
-            apply_affine(t, self.trail), apply_affine(t2, self.location)
+            apply_affine(t2, self.trail), apply_affine(t, self.location)
         )
```
I haven't tested this thoroughly, but this change was needed for [the example in the first comment](https://github.com/danoneata/chalk/pull/102#issuecomment-1214242619) (`dia2`).

chalk/trail.py
```diff
@@ -1,110 +1,166 @@
...
-                                      Defaults to ORIGIN.
-
-        Returns:
-            Path: A path object.
-        """
-        points = [origin]
-        for s in self.offsets:
-            points.append(points[-1] + s)
-        return Path.from_points(points)
+        return Trail(self.segments + other.segments, False)
```
danoneata, line 79: The combination of two trails when one of them is a loop (that is, closed) might not be well defined. I see that in Haskell they have explicitly disallowed the `Monoid` instance for `Trail' Loop v n` and I guess that's one of the reasons why they make the explicit distinction of the two types of `Trail`s (lines and loops) in the type system:
```haskell
-- | The empty trail is constantly the zero vector.  Trails are
--   composed via concatenation.  Note that only lines have a monoid
--   instance (and not loops).
instance (Metric v, OrderedField n) => Monoid (Trail' Line v n) where
  mempty  = emptyLine
  mappend = (<>)
```
I think I'm fine with our approach, but we should at least make a special note in the documentation such that we don't forget about this design choice.

chalk/shapes/path.py
```diff
@@ -1,122 +1,83 @@
...
     @staticmethod
-    def hrule(length: float) -> Path:
-        return Path.from_list_of_tuples([(-length / 2, 0), (length / 2, 0)])
+    def empty() -> Path:
+        return Path([])
 
-    @staticmethod
-    def vrule(length: float) -> Path:
-        return Path.from_list_of_tuples([(0, -length / 2), (0, length / 2)])
+    def __add__(self, other: Path) -> Path:
```
danoneata, line 31: Should we define a `Monoid` abstract base class such that all its subclasses have defined `concat`, as we have discussed in #95?

srush: Yeah, I think we should do that, maybe in a next PR. For now repeating `mconcat` doesn't seem so bad.

chalk/shapes/path.py
srush: Good catch. Fixed.

chalk/trail.py
srush: Nice catch. Fixed.

chalk/trail.py
srush: Oh that's really interesting. I was wondering why they made that type distinction, but this makes a lot of sense. 

In python, I'll just fail if you try to mappend a loop?

chalk/shapes/__init__.py
```diff
@@ -43,34 +32,39 @@ def arc(radius: float, angle0: float, angle1: float) -> Diagram:
...
-    return Path.polygon(sides, radius, to_radians(rotation)).stroke()
+#     Returns:
+#        Diagram
+#     """
+#     return Trail.polygon(sides, radius, to_radians(rotation)).stroke()
 
 
 def regular_polygon(sides: int, side_length: float) -> Diagram:
-    return Path.regular_polygon(sides, side_length).stroke()
+    return Trail.regular_polygon(sides, side_length).stroke().center_xy()
```
danoneata, line 59: For rotations it might be more reasonable to center the polygon in its centroid (using the `centered` method on the trail) instead of the center of the bounding box (using the `center_xy` method on the diagram). That is,
```python
return Trail.regular_polygon(sides, side_length).centered().stroke()
```

chalk/trail.py
danoneata: > In python, I'll just fail if you try to mappend a loop?

Yes, I guess raising an exception when `mappend`ing a loop would be the safest way of avoiding unexpected behavior.

chalk/shapes/path.py
danoneata: Yes, agreed; it's just a line of duplicated code.

danoneata:
I think these are all the comments from my side. I'm very happy with the PR, so feel free to merge it on the main branch!

PR #104, opened 2022-08-22
Fix documentation and a note on internals

danoneata:
I've made a pass through the documentation in order to fix the broken code snippets and examples. In the process I've made some other small changes and, maybe more importantly, I've added [a draft note on the internals of the library](https://chalk-diagrams.github.io/api/internals/); do let me know if there are some other aspects you would like me to touch on or expand.


srush:
Looks great. I like the new internals note.

PR #107, opened 2022-08-25
Fix bugs

srush:
Fixes for 

* https://github.com/chalk-diagrams/chalk/issues/105
* https://github.com/chalk-diagrams/chalk/issues/100

Need these for my project and they work currently. 

Further issues
* Arc trail transforms still don't work exactly right. For instance `reverse` is a bit wonky. should in theory draw the same shape but in the reverse order but doesn't.


srush:
Merging as these aren't particularly interesting.

danoneata:
Sorry, I was away (on holiday) these days and didn't get time to sit and review the PR. I can try to revisit the code once I get back to my normal schedule.

srush:
no worries!

Issue #108, opened 2022-09-07
Reversing arc trails alters their shape


danoneata:
As mentioned in #107 by @srush 

> Arc trail transforms still don't work exactly right. For instance reverse is a bit wonky. should in theory draw the same shape but in the reverse order but doesn't.

PR #112, opened 2022-09-19
Preserve shape of reversed arcs

danoneata:
This PR attempts to address #108. I've added some tests, but they don't really safeguard against the original problem (altering the trail's shape).


srush:
Looks great. Must have missed this!

Issue #113, opened 2022-10-04
Subdiagram Improvements


srush:
Subdiagrams come up a lot more than I thought. But it is a little awkward to work with these at the low level. There are a couple of nice functions (locations, with_name, qualification) that we could add. 

https://diagrams.github.io/doc/manual.html#named-subdiagrams

Relatedly, I would like to be able to have text positioned above an arrow. Not sure the best way to do this, but maybe there is some way to place named control points.

danoneata:
Okay, looking at the first example in their tutorial
```haskell
dia1 = (square 1 # translateY 4 # named "bob" ||| circle 1 # named "joe")
     # connect "bob" "joe"
dia2 = (square 1 # named "bob" # translateY 4 ||| circle 1 # named "joe")
     # connect "bob" "joe"

example :: Diagram B
example = hsep 2 [dia1, dia2]
```
<img src="https://user-images.githubusercontent.com/819256/194569071-9104c7e2-44e2-43d0-9fdd-83c1516eb4e0.png" height=100 />

I see that our implementation has a different behaviour:
```python
from chalk import *

dia1 = square(1).translate(0, -4).named("bob") | circle(1).named("joe")
dia1 = dia1.connect("bob", "joe")

dia2 = square(1).named("bob").translate(0, -4) | circle(1).named("joe")
dia2 = dia2.connect("bob", "joe")

dia = hcat([dia1, dia2], sep=2)
```

![o](https://user-images.githubusercontent.com/819256/194569796-aeb4a99f-d5f0-4d11-8e37-f94cae6e88cb.png)

I'll need to check what's going on.

srush:
This seems very subtle. At first read I had it backwards what the behavior would be. It's quite counter intuitive to me that 

```haskell
square 1 # translateY 4 # named "bob"
```

Would connect to the old value? Is it because of the way their semantics work? It seems from their hierarchy names propagate like style.  

https://hackage.haskell.org/package/diagrams-core-1.5.0.1/docs/Diagrams-Core-Types.html#g:4

srush:
Oh I see the problem. We are using `envelope.center` in connect which is an approximation of the midpoint along the x/y axes. They are using the true local origin (0, 0) of the subdiagram. We actually don't have access to that in our code. 

Proposal: 

* Remove `get_subdiagram_trace` and `get_subdiagram_envelope`
* Add `get_subdiagram`
[...]

PR #114, opened 2022-10-04
Two SVG bugs

srush:
* Newlines \n in svg specification seems to break in some browsers. 
* Images that are too large and then scaled down cause Affine to think transformation are degenerate. 

Small temporary fixes.


danoneata:
Hey! Thanks for the fixes!

The image fix works only for the SVG backend, right? I wonder if it would be more flexible (backend agnostic) to apply the scaling only when [we create the image primitive](https://github.com/chalk-diagrams/chalk/pull/114/files#diff-3e9c25ae25ab46ed4c6e9716d4de8f9facac02c3390ba49756826837219b0c39R64):
```diff
 def image(local_path: str, url_path: Optional[str]) -> Diagram:
     from chalk.core import Primitive

-    return Primitive.from_shape(Image(local_path, url_path))
+    return Primitive.from_shape(Image(local_path, url_path)).scale(0.05)
```

Somewhat unrelated, but what should the `url_path` argument from the `image` function be set to (I see that's used  only in the SVG backend)?

srush:
I think we need a couple long term fixes. 

1) figure out why large scaling ops break the degenerate check in affine.
2) make `image` backend specific. Diagrams also has some backend specific code (svgtext)
3) pull out and document scaling constants. I.e default images to local width 1? Not sure what is right to do here.

danoneata:
Okay! So I guess you are suggesting to merge these temporary fixes for the time being and create separate issues for the long-term problems, right? I'm fine with that, but I just wanted to point out that the current changes will affect the image rendering with `cairo` (since the `Image`'s `width` and `height` do not match the true `width` and `height` of the PIL image data); I don't have a satisfactory solution—maybe we could try resizing the image by the 0.05 factor in the `from_pil` function?
```python
     format: cairo.Format = cairo.FORMAT_ARGB32
+    im = im.resize((int(im.width * 0.05), int(im.height * 0.05)))
     if "A" not in im.getbands():
```

srush:
We don't need to merge. I'll fix Cairo for real here. 

Mostly just sent this to remind myself since I needed it in my code.

danoneata:
Ah, cool! Sorry for misunderstanding! I'll take a look at #113 then if you haven't started already working on it.

PR #116, opened 2022-10-07
Subdiagram improvements

danoneata:
This PR addresses #113, but it's still work in progress. I've started by fixing the example from that issue and by following [the plan laid out by Sasha](https://github.com/chalk-diagrams/chalk/issues/113#issuecomment-1271741996). The next steps are to add the extra functionality.

TODO
- [x] `get_location` to retrieve the local origin of the `Subdiagram`
- [x] `qualify` diagram names by prefixing them with a given name
- [x] `with_name` (unclear whether this function can be used easily in Python, given that its most natural style involves nested lambda expressions) 
- [x] `boundary_from` to return points on the boundary of a given `Subdiagram` given a direction vector
- [x] `show_labels` to show label names on top the diagram
- [x] Documentation

Deferred sub-tasks:
- `style` as an attribute to the `Subdiagram` class (since it was unclear to me whether this is useful for the current API)
- Draw labels on arrows (since it would be better served by a different PR)


srush:
I agree withName is a bit too aggressive for python. It looks like there is also a with_names function though that might be better.

danoneata:
> It looks like there is also a with_names function though that might be better.

Yes, you are right, `with_names` looks more appropriate for us. I've replaced `with_name` with `with_names` and added some examples (translated from Haskell) in [the documentation](https://chalk-diagrams.github.io/api/names/).

I think I'll move the task of adding labels to arrows to a separate issue (and PR) since it appears a bit more involved at first sight, especially if we want it to work on arrows with arbitrary trails (shafts).

chalk/core.py
```diff
@@ -62,15 +63,17 @@ def compose(
     ) -> Diagram:
         return Compose(envelope, self, other if other is not None else Empty())
 
-    def named(self, name: str) -> Diagram:
+    def named(self, name: Union[str, Tuple[str]]) -> Diagram:
```
srush, line 66: I think Tuple[str] is only unary tuples. Do you want this be arbitrary tuples?

chalk/subdiagram.py
```diff
@@ -20,27 +22,39 @@
 
 
 Ident = Affine.identity()
-Subdiagram = Tuple[Diagram, Affine]
+Name = Tuple[str, ...]
```
srush, line 25: I have been using non-string names. Maybe we can make named() take Any but then wrap in a `Name` dataclass?

srush: I guess the question is do we expose Name to the user. Haskell does, but it forces you to use .> to build compound names.

chalk/subdiagram.py
```diff
@@ -87,12 +101,92 @@ def visit_apply_name(
...
+    if any(sub is None for sub in subs):
+        return self
+    else:
+        return f(subs, self)
+
+
+SubMap = Dict[Name, List[Subdiagram]]
+
+
+class GetSubMap(DiagramVisitor[SubMap]):
```
srush, line 136: This is nice! good use of Visitor.

chalk/core.py
```diff
@@ -270,8 +283,61 @@ def accept(self, visitor: DiagramVisitor[A], **kwargs: Any) -> A:
...
 
-    dname: str
+    dname: Name
     diagram: Diagram
 
     def accept(self, visitor: DiagramVisitor[A], **kwargs: Any) -> A:
         return visitor.visit_apply_name(self, **kwargs)
+
+
+class Qualify(DiagramVisitor[Diagram]):
```
srush, line 293: Interesting. I was thinking we could do this by having a "ApplyQualifier" node, that gets checked by the other visitor. Changing the tree scares me a little bit.

chalk/arrow.py
```diff
@@ -30,27 +30,42 @@ class ArrowOpts:
 def connect(
     self: Diagram, name1: str, name2: str, style: ArrowOpts = ArrowOpts()
 ) -> Diagram:
-    bb1 = self.get_subdiagram_envelope(name1)
-    bb2 = self.get_subdiagram_envelope(name2)
-    return self + arrow_between(bb1.center, bb2.center, style)
+    sub1 = self.get_subdiagram(name1)
```
srush, line 33: Might consider doing these with withNames? Don't think the code would change much, but it would be a good test.

chalk/core.py
danoneata: Thanks a lot for the feedback! Yes, I'll correct this and run `mypy` as well (I've postponed running it since I think its latest version (?) shows errors in multiple places in the codebase).

chalk/core.py
danoneata: Okay, I can add a new `ApplyQualifier` variant to our abstract syntax tree (AST). I think my approach was a reflex to keep the core AST minimal, but indeed traversing the AST when running `qualify` might incur too much computational burden.

chalk/subdiagram.py
danoneata: Sure, I will use more generic types for the atomic names (in principle, I think that anything that's hashable should work, since we index the subdiagram dictionary by a `Name`)?

Regarding the second point, do you have any advice? I've tried to keep things simple and allow the user to either specify an atomic name (a string in the initial version of the code) or a compound / qualified name (a tuple of strings); but I'm open to suggestions.

chalk/core.py
srush: I trust your judgement.

chalk/subdiagram.py
srush: Hmm. this is actually a little tricky. I think a tuple should work for qualification. But I have been using Tuples for just standard names -> i.e. (i,j) for a cell in a matrix. The question is will that break?

danoneata: Yes, you are right, I think my approach wouldn't have worked for that case.

As a second version, I was thinking about something along these lines:
```python
AtomicName = Any
Name = Tuple[AtomicName, ...]

class ApplyName(BaseDiagram):
    dname: Name
    diagram: Diagram

def named(self, *name: AtomicName) -> Diagram:
    return ApplyName(name, self)

def qualify(self, *name: AtomicName) -> Diagram:
    return self.accept(Qualify(name))

def get_subdiagram(self: Diagram, *name: AtomicName) -> Optional[Subdiagram]:
    return self.accept(GetSubdiagram(name), t=Ident)

def with_names(
    self: Diagram,
    names: List[Name],
    f: Callable[[List[Subdiagram], Diagram], Diagram],
) -> Diagram:
   ...
```
which would allow for something like this
```python
s = square(1)
squares = (
    (s.named("N", "W") | s.named("N", "E")) /
    (s.named("S", "W") | s.named("S", "E")))
squares = squares.qualify(0)
squares.get_subdiagram(0, "N", "W")
```
But unfortunately the API for `with_names` and `connect` get confusing since the name arguments have to be `Tuple`s and not `AtomicName`s (which in essence means that the user needs to know that the `Name`s are represented as `Tuple`s internally...):
```
root = circle(1).named("root")
leaves = hcat([circle(1).named(c) for c in "abcde"], sep=0.5).center()
nodes = root / vstrut(2) / leaves

def connect(subs, nodes):
    ...

for c in "abcde":
    nodes = nodes.with_names([("root", ), (c, )], connect)
```

srush: Yes, I agree this a bit confusing. My intuition is that we should have the correct, verbose way first. There are a couple areas in the code that we may want "abbreviations" as you have mentioned in the past (colors, P2s, etc). Maybe we just have ways to make it less verbose on top of this foundation.

danoneata: Thanks a lot for the feedback @srush! I've updated the API to be more explicit: all naming functions (`named`, `qualify`, `with_names`, `get_subdiagram`, etc.) are now expected to take as input an instance of the `Name` class. There are some examples in the [`examples/subdiagrams.py`](https://github.com/chalk-diagrams/chalk/blob/8e4e5ab738450aee41257d1b4bdee460204a064f/examples/subdiagrams.py) file. When you get a chance, please let me know what you think of the current implementation. I'll update the other examples (`lenet.py`, `lattice.py`, `tree.py`, etc.) and the documentation, once we converge to a solution.

PR #127, opened 2024-02-12
Fix link to Escher example

BioGeek:
Fixes the link to [Escher’s  “Square Limit”](https://github.com/chalk-diagrams/chalk/blob/master/examples/escher_square_limit.py) example in the README.


danoneata:
Thanks for this fix, as well!

PR #130, opened 2024-02-29
Isometric Example

srush:
Was playing around with 3d and thought this was a fun example based on the haskell isometric projection. In theory this could be a transform3d, but probably not worth the effort. 

What might be fun is to replace transform with numpy generally... That would allow very fast scaling of trails,.


danoneata:
Oh, neat! I should play with this sometime! I'll add the NumPy-based transform to the [TODO list](https://github.com/chalk-diagrams/chalk/issues/113#issuecomment-1299205037) 🙂 

Talking about other directions to investigate, I was thinking about animations. Would that be an useful feature to have? And if so, how difficult would it be to implement?

srush:
Animations would be really cool... I'm a bit worried it would be really hard though. I have no idea how to render them, or if there is a way to dynamically update only part of the tree. It's also kind of annoying that SVG doesn't really support animations naturally. (I see there are some ideas in haskell though. )

This is Chalk, but I just rendered a lot of images and gif'd them together-> 
https://twitter.com/i/status/1757031261556629911

danoneata:
Yes, your Drop7 demo might make a nice motivating example. Although it does look rather nice as it is 🙂 

I see that in Haskell the idea for animations is quite simple: an animation is a time-varying diagram, implemented as a function from `Time` to `Diagram` (plus information regarding the start and end times). Even simple, this abstraction lends itself to many combinators, in particular, it is an `Applicative`. I'm not sure about its efficiency though: it doesn't look that there is any explicit sharing done across time. In terms of inspiration, it might also be worth taking a look at 3blue1brown's [manim](https://github.com/3b1b/manim) library. 

I'm currently caught up with some deadlines, but I might be able to think about this in mid-March or so.

PR #132, opened 2024-05-14
Rewrite and simplification of Chalk core

srush:
I was reading about some of the more recent updates to the Python type system and thought I would try a new version of the chalk core that fixed things that bothered me. This is a rather big PR, but most of it is pretty simple. 

Changes: 

* Changed the Compose node to have a list of children instead of 2. 
* Changed + to merge to the list instead of adding to the tree when possible. 
* Added a Monoid "typeclass" with a default + and empty, and a general concat. Nearly everything in chalk is a monoid now
* Added a Maybe / MList to make these monoid'able. 
* Make Visitor define an explicit generic argument for its down / up elements. Default behavior is that the down element is a monoid and define default behavoirs for compose, empty, etc. Remove kwargs hack.  
* Replace fake placeholder TTrans types with new Self argument. 
* Remove repeated Abstract class / Interface overlaps. (this is probably incorrect, but it makes the code simpler)

Issues: 

* Currently it works with python 3.10+. Need to try with earlier versions.

Why?

* Reason 1 - cleaner types and less code to simplify things. 
* Reason 2 - I've been playing around with this library Penzai that includes a really nice tree visualizer. with a couple of lines of code you can view the whole tree from Chalk when you draw. It's pretty slick. 

https://chalk-diagrams.github.io/viz/

image


danoneata:
Thank you, Sasha! I'll take a closer look at the PR in the next couple of days (on Thursday probably). The penzai visualization looks sleek! I think it can be really useful for debugging!

chalk/core.py
```diff
@@ -63,7 +69,14 @@ def _style(self, style: Style) -> Diagram:
...
     ) -> Diagram:
-        return Compose(envelope, self, other if other is not None else Empty())
+        if isinstance(self, Compose):
+            if isinstance(other, Compose) and other is not None:
+                return Compose(envelope, self.diagrams + other.diagrams)
+            if other is not None:
+                return Compose(envelope, list(self.diagrams) + [other])
+        return Compose(
+            envelope, [self, other if other is not None else Empty()]
+        )
```
danoneata, line 79: The logic here might be a bit difficult to follow (at least for me). For example, it seems that we are missing a case: when `self` is not `Compose`, but `other` is `Compose`. For this case, we should concatenate `self` wrapped as list with the diagrams unpacked from `other`.

What about something along these lines:
```python
other = other if other is not None else Empty()
if isinstance(self, Compose) and isinstance(other, Compose):
    return Compose(envelope, self.diagrams + other.diagrams)
elif isinstance(self, Compose):
    return Compose(envelope, self.diagrams + [other])
elif isinstance(other, Compose):
    return Compose(envelope, [self] + other.diagrams)
else:
    return Compose(envelope, [self, other])
```

Or with pattern matching (but this requires Python 3.10+):
```python
other = other if other is not None else Empty()
match (self, other):
    case (Compose(_, ds1), Compose(_, ds2)):
        return Compose(envelope, ds1 + ds2)
    case (Compose(_, ds1), _):
        return Compose(envelope, ds1 + [other])
    case (_, Compose(_, ds2)):
        return Compose(envelope, [self] + ds2)
    case (_, _):
        return Compose(envelope, [self, other])
```

chalk/visitor.py
```diff
@@ -14,45 +14,63 @@
...
 
-    def visit_empty(self, diagram: Empty) -> A:
-        raise NotImplementedError
+class DiagramVisitor(Generic[A, B]):
+    A_type: type[A]
 
-    def visit_compose(self, diagram: Compose) -> A:
-        raise NotImplementedError
+    def visit_primitive(self, diagram: Primitive, arg: B) -> A:
+        "Primitive detauls to empty"
```
danoneata, line 30: Typo: "defaults" 🙂

danoneata:
> - Make Visitor define an explicit generic argument for its down / up elements.

Nice; I like that we now have the input type explicitly specified for the `visit` methods!

>  Default behavior is that the down element is a monoid and define default behavoirs for compose, empty, etc.

I was curious: for the `ToSVG` visitor, the return type is `svgwrite.base.BaseElement`; how does that the type system know that this is a `Monoid`?

By the way, how do you type check code? Are you still using `mypy`? Which version and which options?

srush:
* Oh the BaseElement thing might be a bug. I think MyPy doesn't have the stubs for `svgwrite` and so it is likely failing to check that.

*  I am using mypy 1.10, should we switch to pyright?

danoneata:
Thanks for the pull request and merge!

> I am using mypy 1.10, should we switch to pyright?

Hmm... I don't have a strong opinion towards one or the other. But I see that pyright reports a few errors on our codebase (although I expect these to be fixable).

srush:
K, I think I'll fix these, bump the required version to 3.10+, update the docs, and then up the version number. Excited to check out the animation lib as well.

